PR #32, opened 2016-02-08
Use relative imports instead of sys.path monkeypatching.

rec:


rec:
Let me know if you want tips as to how to avoid those unnecessary "merge commits" when you submit.

adammhaile:
Huh... weird. I could not get that working in the beginning. Will play with
it. Not sure what I did wrong... thanks!

rec:
Oh, I've seen that antipattern before, it comes from not knowing relative imports _and_ having files at different levels.  In that case, there is no way to get to the symbol at the top level...

The reason I care is that it adds an element into your `sys.path` every time you reload _any_ of these individual files.  So if you load the library a couple of times for various reasons, your `sys.path` becomes monstrously huge.  :-)

Issue #35, opened 2016-02-19
un-Pythonic weirdness in `bibliopixel.util.d`


rec:
https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/util.py#L1-L18

You very rarely have to use `__dict__` in Python, and I can't even understand what this is supposed to do, but I'm about 98% sure that there is some way to do this in Python with a couple of lines of code with no `__` in them.  :-D

adammhaile:
LOL... That's used in PixelWeb extensively. It makes all dictionary keys accessible as attributes.
myDict.hello instead of myDict['hello']

And yes, it's a total hack. I'm all ears if there's a more pythonic way to do this.

adammhaile:
Fixed with pull request #40

PR #40, opened 2016-02-21
Simplify util.d (#35).

rec:
You don't use `update` here but you do in PixelWeb - once.

I'd kill it entirely and replace it with, well, nothing :-D but just a constructor to `AttributeDict` (aka `d`)


adammhaile:
Hahaha... is it seriously that easy?

Issue #44, opened 2016-03-23
Python 3 support?


rec:
I'm going to file this, even if we close it as "will not fix", just to have a record.

Basically, it isn't too hard to change a Python program to make it support both 2 and 3, using [six](https://pypi.python.org/pypi/six) - a single file solution.

Also, this task can be done "piece at a time" - you don't have to do it all it once or even all of one file at a time.  Of course, you can't fully test it until it's all converted.

The reason to do this is because as a "patriotic" Python developer, I feel bad doing new development in 2 only.

adammhaile:
Agreed. This will be a will not fix for master/dev, but we can tag this as
v3.0 branch only. Will be good to have. The lack of 3 support was orginally
because some dependencies didn't support it, but that's not such an issue
anymore. The main culprit was py-spidev, but v3.0 supports python 3, so
we're good.

rec:
We can support both 2 and 3 in the same codebase, though, so this isn't really a breaking change!

adammhaile:
I realize... but at this point the v2 branch is stable and we have so many
changes planned for v3, that I'd rather just put anything that's not bug
fixes to v2 into the new v3 branch for future release. I might through a
couple new changes into v2 related to some other projects I'm working on,
but only because those projects won't want to wait for v3 to be stable.

PR #47, opened 2016-05-11
Use absolute import paths everywhere.

rec:
* Prepare for Python 3.x.
* Works in Python 2.7.

I was actually able to run a test script with a visualizer only with this under Python 3!  Further testing is required...

### DIFF
--- a/bibliopixel/animation.py
+++ b/bibliopixel/animation.py
@@ -1,14 +1,8 @@
-import time
-import log
+import threading, time
 
-from led import LEDMatrix
-from led import LEDStrip
-from led import LEDCircle
-import colors
-
-from util import d
-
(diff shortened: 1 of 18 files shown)


rec:
Hmm, I'm now not 100% sure that this works right in Python 2.7 (my testing was wrong).

I'll check it out in the next day or so (I know you're traveling so I'm not feeling rushed...)

rec:
OK, I just pushed a rebased change that does run on Python 2.

The issue is that visualizer.py is a file that runs at a different base path from every other .py in the system.  I kept most of the code where it was, but I made a tiny file at the top that calls that code for the visualizer.

rec:
#44

PR #55, opened 2016-08-02
Fast gamma and permutation calculations

rec:
Finally!  :-)

I set it up so it will work with the older version of timedata, but in order to get the fancy rendering you probably need to update to the current master.

I introduced a long-named method that later vanishes once I figured it all out, but not willing to spend the time to rebase and make it beautiful - each step is correct, if perhaps not lovely.


rec:
I pushed that last commit on to deal with masterBrightness - but really, what's there might not even work!

I think I need to add a bit more to only use it - if it's set - in the very last stage (and perhaps with timedata...)

rec:
to be clear - what I mean is that I didn't change the brokenness or lack thereof.  :-)

PR #67, opened 2016-10-08
Added baudrate to constructor

tom-slick:


bibliopixel/drivers/serial_driver.py
```diff
@@ -88,12 +88,21 @@ class DriverSerial(DriverBase):
...
 
-    def __init__(self, type, num, dev="", c_order=ChannelOrder.RGB, SPISpeed=2, gamma=None, restart_timeout=3, deviceID=None, hardwareID="1D50:60AB"):
+    def __init__(self, type, num,
+                 dev="",
+                 c_order=ChannelOrder.RGB,
+                 SPISpeed=2,
+                 gamma=None,
+                 restart_timeout=3,
+                 deviceID=None,
+                 hardwareID="1D50:60AB",
```
rec: `"1D50:60AB"` ??  :-D

Yes, I know you didn't do that.  @adammhaile , what's the scoop here?

bibliopixel/drivers/serial_driver.py
```diff
@@ -248,13 +257,12 @@ def _generateHeader(cmd, size):
...
 
-    @staticmethod
-    def setDeviceID(dev, id):
+    def setDeviceID(self, dev, id):
         if id < 0 or id > 255:
             raise ValueError("ID must be an unsigned byte!")
 
         try:
-            com = serial.Serial(dev, timeout=5)
+            com = serial.Serial(dev, baudrate=self.baudrate, timeout=5)
```
rec: If we're going to bring in the baudrate as a parameter, we should also bring in the timeout as well conceivably.

bibliopixel/drivers/serial_driver.py
```diff
@@ -88,12 +88,21 @@ class DriverSerial(DriverBase):
     deviceIDS = {}
     deviceVers = []
 
-    def __init__(self, type, num, dev="", c_order=ChannelOrder.RGB, SPISpeed=2, gamma=None, restart_timeout=3, deviceID=None, hardwareID="1D50:60AB"):
+    def __init__(self, type, num,
```
rec: +1 for cleaning up the parameter list.  :-D

adammhaile:
Will respond more later. The hardwareID is for the AllPixel for which we
own a USB ID. It allows connection without specifying the port. We default
it to this because most people use the AllPixel with this so it is easier
for them. But setting "dev" overrides this.

On Oct 8, 2016 11:53 AM, "Tom Swirly" <notifications@github.com> wrote:

*@rec* commented on this pull request.

Thanks for doing this! :-)

My theory is that this isn't urgent for you to have on master, so we'll
bring in this change - or a change much like it - onto the dev branch which
I am sitting on (the actual dev/3.0 branch is sorta broken now, but I have
a fixed one).

I'd rather not bring it onto master right now because then we have to
rebase all of dev.

I have no problems keeping this pull request open as a reminder until then!
------------------------------

In bibliopixel/drivers/serial_driver.py
<https://github.com/ManiacalLabs/BiblioPixel/pull/67#pullrequestreview-3410277>
:

> @@ -88,12 +88,21 @@ class DriverSerial(DriverBase):
     deviceIDS = {}
     deviceVers = []

-    def __init__(self, type, num, dev="", c_order=ChannelOrder.RGB,
SPISpeed=2, gamma=None, restart_timeout=3, deviceID=None,
hardwareID="1D50:60AB"):
+    def __init__(self, type, num,
+                 dev="",
+                 c_order=ChannelOrder.RGB,
+                 SPISpeed=2,
+                 gamma=None,
+                 restart_timeout=3,
+                 deviceID=None,
+                 hardwareID="1D50:60AB",

"1D50:60AB" ?? :-D

Yes, I know you didn't do that. @adammhaile <https://github.com/adammhaile>
, what's the scoop here?
------------------------------

In bibliopixel/drivers/serial_driver.py
<https://github.com/ManiacalLabs/BiblioPixel/pull/67#pullrequestreview-3410277>
:

>          if id < 0 or id > 255:
             raise ValueError("ID must be an unsigned byte!")

         try:
-            com = serial.Serial(dev, timeout=5)
+            com = serial.Serial(dev, baudrate=self.baudrate, timeout=5)

If we're going to bring in the baudrate as a parameter, we should also
bring in the timeout as well conceivably.
------------------------------

In bibliopixel/drivers/serial_driver.py
<https://github.com/ManiacalLabs/BiblioPixel/pull/67#pullrequestreview-3410277>
:

> @@ -88,12 +88,21 @@ class DriverSerial(DriverBase):
     deviceIDS = {}
     deviceVers = []

-    def __init__(self, type, num, dev="", c_order=ChannelOrder.RGB,
SPISpeed=2, gamma=None, restart_timeout=3, deviceID=None,
hardwareID="1D50:60AB"):
+    def __init__(self, type, num,

+1 for cleaning up the parameter list. :-D

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/ManiacalLabs/BiblioPixel/pull/67#pullrequestreview-3410277>,
or mute the thread

rec:
Got it!

I shall extract this as a named constant at some point so the next guy won't have to ask the same question.

Have a great vacation!

tom-slick:
I think I found a problem.  I only tested it with ledstrips hooked up.  I had a friend try it out on an AllPixel and got an error when testing with LEDMatrix      
TypeError: unbound method getDeviceID() must be called with DriverSerial instance as first argument (got str instance instead)﻿

We got it working by adding the baudrate as class level variable.  So It seems this will need a rethink.

Close as has bugs.  I still think this is worth adding to the code, just not this way.

rec:
Well, gosh, thanks for finding the bug before we even really looked at it!

I'll file a bug and we'll get to this - actually, we'll get to it pretty fast as we're hell-bent on a release in the next few weeks.

PR #72, opened 2016-12-16
Split led.py into multiple files.

rec:
The API should be identical for client code!


adammhaile:
Nice! :) Glad I did the font stuff already... Looks great :) 
+2
Note... I've disabled the create a merge commit and squash and merge options. You should be able to choose rebase and merge and it will do what we've discussed previously with no dirty merge commits. I've done some testing in another repo and GitHub will automatically disallow this option if there would be any weirdness or errors... so if there were, then we could use the instructions you previously sent, but this should work fine assuming that the PRs are not in conflict... as in the PR is based on a branch that already includes all commits currently in the branch being pushed to.
Anyways... that's my opinion, I'm up for rebuttal on the usage of the above "Rebase and merge" button.

rec:
That seemed to work!  :-)

adammhaile:
Awesome :)

PR #73, opened 2016-12-18
New util subdirectory;  split up animation.py.

rec:
I didn't try to be fancy here, and you might well decide that there are two many small files in `animation/` and later want to glue them together, but as this is, it's pretty understandable and uncontroversial.


rec:
Tried the "request review" feature just to see what it did.  :-D

rec:
Ah, thought I'd routed out all uses of d!  I'll put it back, moment pls.

If you recall, you had some complex thingie for `d` which did exactly what `AttributeDict` does.  I gave it that name because `d` is a pathologically bad name for a class!

rec:
I just removed that commit and rebased - no sweat!

adammhaile:
@rec - LOL, yeah, I'm lazy so I like short names so I can type them faster :P Though my main concern is that d() did more than AttributeDict, for example iterrating through the dict and converting everything below it into the same. Otherwise. myDict.name might work but myDict.name.last would have to be myDict.name['last']. Or am I missing something and AttributeDict will do the same?

rec:
Well, what was there before Feb 21 https://github.com/ManiacalLabs/BiblioPixel/commit/9542735e41135e7e830e252f318f84197fcbf31b didn't do that right.  :-D  It only did it for two levels!

I can easily add that change separately - very reasonable ide... but that's independent of this change.

adammhaile:
Oh yeah... didn't realize that d() had basically been gone forever. Whoops.
+1 to all of this. Will merge unless you've got more.

rec:
I did change the meaning before, however!

Overall, either `d` or `AttributeDict` are both broken because if you assign something after it's created, you get bad results.

Let me fix that - in a separate change...

Good to go!

PR #74, opened 2016-12-18
importer.py imports modules or symbols.

rec:
As you know, I want to be able to load a configuration of BP from a text file, and this is the first step.


adammhaile:
Can you give an example usage of this? Want to make sure we agree on the direct (config format, etc.) this is going...

rec:
Well, to be honest, I don't know quite what it is yet :-D but I do know that it will have to use this method to load classes in.

adammhaile:
Ha... well, ok. My main concern is whatever this text file format is. Let's
not re-invent and jetpacks here. JSON is built into python and works quite
well. Plus PixelWeb already uses that. Personally I'd rather something with
less curly brackets, but that adds further dependencies. Such as yaml...
nice and terse like Python but you have to install pyyaml or similar.

rec:
My thought was entirely to use the constructors that already exist, and just store the parameters as data.

This would make the format independent of BiblioPixel.

I need this work anyway for "startuppy" - perhaps I should put it there and bring it over?  But for now I just want to be able to switch between drivers easily.

adammhaile:
Interesting... well, still would like to see what the text file format is.

rec:
You'll be the first know - but no matter what, it'll need to dynamically load classes from modules!

adammhaile:
K. +1

PR #76, opened 2016-12-18
Restoring brightness control for hardware devices.

adammhaile:


rec:
First comment - this is two separate commits in one, struggling to get out!

Could I convince you to extract out the receiver as a separate pull request?

More comments coming on "brightness".

bibliopixel/drivers/driver_base.py
```diff
@@ -97,7 +97,7 @@ def set_brightness(self, brightness):
         if brightness > 255 or brightness < 0:
             raise ValueError('Brightness not between 0 and 255: %s' % brightness)
         self._brightness = brightness
-        return True
+        return False  # Device does NOT support internal brightness control
```
rec: You don't need this comment here.  Instead, you should document the method and explain what the return value means.  If you in fact continue to keep this method like this.

--

Honestly, though - is "brightness" really a concept we want to continue to support in exactly this way? 

I believe that your planned use of "brightness" is a way to "tune" drivers where some strips are perceptually darker or brighter than others.  This means that you are expecting to set and forget this value - or at least go through a tuning phase and then leave it alone.

Great!  Very reasonable _goal_ - it's the _mechanism_ I'm disputing, for several reasons:

* There are now two sorts of things you can set in a driver:
  * things you set at construction, like the gamma
  * brightness.

* It involves two additional per pixel computations - a multiply and a divide - on top of the gamma.
* Using 0-255 is pretty unintuitive when you're talking about brightness!

---

[much thought there...]

 I think the driver class needs to split!

I think we need a "dumb data" driver info class that contains:

* the driver type
* gamma (default comes form the driver)
* offset (default comes from the driver)
* rgb permutation
* brightness
* columns
* rows
* rotation
* reflection
* scan alternate rows

Some of this is already in the matrix class and I think this isn't right... we might want code as well to do these as special effects in the signal chain.
 
There needs to be an Impl implementation class for each type of drive (LPD8806, Hue, etc) which has _no_ data on it - works on the "dumb data" driver class.

rec:
That convinced me!

Glad to work out how it should work, maybe. Maybe next time!  :-D

If you cut out the receiver part into a separate change _and_ give it a better commit message, I'll +1 it.

"fixing brightness for devices, adding receiver in"

has some "quality defects", particularly the intriguing gap after "in".  :-P

SINCE I'm whining about this anyway, here's the typical rules for a commit message.

* Starts with a capital, ends with a . (i.e. a full  English sentence).
* In the active present voice (i.e. the simplest tense possible of the verb! no -ing or -s or -ed).
* Only has one thing in it!

So this becomes "Fix brightness for devices." and receiver goes into its own commit!

adammhaile:
You sound like my manager...

adammhaile:
Changed. Though I'm leaving the comment in, at least for now. Personally I think there's not nearly enough comments in the code. Yes, good docs are needed, but at the same time, inline comments makes sense. If they had been there in the first place, this wouldn't have been broken ;)

rec:
Well, I was wanting you to document the return type!  So it gets into the automatically generated docs we will one day make.  :-D

Merging!

rec:
Oh, and I hadn't worked out that brightness was "everything"!  Now I need to think some more.  But for now - food!

adammhaile:
woot :)

adammhaile:
LOL... I figured as much ;)

PR #79, opened 2016-12-19
Only report timedata status on startup in debug.

rec:
Gets rid of one spammy line in the build/test output.  :-D


adammhaile:
+1

PR #80, opened 2016-12-23
Simplifications to animation/base.py.

rec:
Some tweaks to clear things up.  No external APIs change _except_ `untilComplete` becoming `until_complete`.

I was going to forestall some minor objections I expect you'll have, but let's wait for the review.

I toyed with refactoring BaseAnimation.run more dramatically - with [this](https://github.com/rec/BiblioPixel/commit/43a4a61cdf584bfeb99cc0f85505ff5d5bfd6190) and [this](https://github.com/rec/BiblioPixel/commit/086b241d64d585da1e05db6be28c488261606dba) - but I feel this isn't really fixing the underlying problems, though it "looks" better, so these aren't on here.


bibliopixel/animation/base.py
```diff
@@ -56,21 +34,29 @@ def stopThread(self, wait=False):
...
+
+    def cleanup(self):
         self.stopThread(wait=True)
         self._led.all_off()
         self._led.push_to_driver()
         self._led.waitForUpdate()
 
-    def cleanup(self):
-        return self.__exit__(None, None, None)
+    def run_cleanup(self, **kwds):
```
adammhaile: where does this actually get called?

bibliopixel/animation/base.py
```diff
@@ -133,55 +117,60 @@ def _run(self, amt, fps, sleep, max_steps, untilComplete, max_cycles, seconds):
...
-            args = {}
-            l = locals()
-            run_params = ["amt", "fps", "sleep",
-                          "max_steps", "untilComplete", "max_cycles", "seconds"]
-            for p in run_params:
-                if p in l:
-                    args[p] = l[p]
+        def do_run():
+            self._run(
+                amt, fps, sleep, max_steps, until_complete, max_cycles, seconds)
```
adammhaile: not a fan of this newline

bibliopixel/animation/base.py
rec: It's now gone in the later changes I pushed - because `run` cleaned up anyway, except when there was an exception, and I fixed that second part.

bibliopixel/animation/base.py
rec: yes, looks bad... let me do something... Hah, I must have not liked it either, because it's cleaned up in a later commit!

adammhaile:
+1

PR #81, opened 2016-12-23
util.cleaner is a context manager that calls the cleanup() method.

rec:
See https://groups.google.com/forum/#!topic/bibliopixel-dev/Hacj4vQFjZM


adammhaile:
+1

PR #82, opened 2016-12-23
Improve testing and benchmarks in setup.py.

rec:
Remove duplicate code;  stop execution if there's an error and propagate the error code back to the shell, so you can detect that the tests failed in your scripts.


rec:
Oh, and note that there's an niggling little inconsistency - it's `setup.py test` but `setup.py benchmarks` (with an s).  Considering quickly changing that - I'd prefer `tests` but `test` is sort of built-in, so what about making it `benchmark`?

adammhaile:
Agreed. Let's make it benchmark. I would name test tests, but agreed that
test is the more common standard.

rec:
Done.

PR #86, opened 2016-12-28
upped to 3.0 version

adammhaile:


rec:
For Christmas, I should have sent you a Shift key and a period!  :-P

PR #87, opened 2016-12-29
DriverSerial now accepts baudrate parameter.

adammhaile:


bibliopixel/drivers/serial_driver.py
```diff
@@ -80,7 +80,11 @@ class DriverSerial(DriverBase):
     deviceIDS = {}
     deviceVers = []
 
-    def __init__(self, type, num, dev="", c_order=ChannelOrder.RGB, SPISpeed=2, gamma=None, restart_timeout=3, deviceID=None, hardwareID="1D50:60AB"):
+    def __init__(self, type, num, dev="",
+                 c_order=ChannelOrder.RGB, SPISpeed=2,
+                 gamma=None, restart_timeout=3,
+                 deviceID=None, hardwareID="1D50:60AB",
```
rec: Consider extracting `'1D50:60AB'`as a constant!

bibliopixel/drivers/serial_driver.py
```diff
@@ -80,7 +80,11 @@ class DriverSerial(DriverBase):
...
+                 baudrate=921600):
```
rec: Definitely extract `921600` as a constant, because it appears three times, and because it really isn't clear where it comes from - aside from being 900 * 1024.

bibliopixel/drivers/serial_driver.py
```diff
@@ -220,12 +224,12 @@ def _connect(self):
             raise e
 
     @staticmethod
-    def setDeviceID(dev, id):
+    def setDeviceID(dev, id, baudrate=921600):
```
rec: Eventually, these could just be pure functions and not static methods, but that's for another day.  :-)

bibliopixel/drivers/serial_driver.py
adammhaile: 921600 is a standard baud rate.

bibliopixel/drivers/serial_driver.py
adammhaile: Probably, but yeah... not the time for that.

bibliopixel/drivers/serial_driver.py
rec: Hmm, a little baffled as to what you mean, but let's push this anyway.

bibliopixel/drivers/serial_driver.py
rec: Clearly not so standard that I knew about it!

This is very un-[DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) but we can do this later.  Merging!

adammhaile:
Hey... We said the committer was to merge!

bibliopixel/drivers/serial_driver.py
adammhaile: What you are remembering are old RS-232 baud rates... 921600 is very standard for anything that's USB Serial. I use it all the time and it's the default for the ATMega32u4 USB stack that I use in the AllPixel.

adammhaile:
LOL... I believe we did :P

PR #88, opened 2017-01-01
Some first layout code.

rec:
Unfortunately, this isn't yet advanced enough to replace existing code, but it does have tests.

A `Strip` is like a very stripped down, fixed-length `list`.  In fact, you can always use a `list` when you need a `Strip`.

`Segment` is what we're going to be carving out of our main `ColorList` to give to `Device`s and `Layouts`.  Generally, our "lists" only have to support `len(x)`, `x[i]` and `x[i] = value`, where `i` is only a simple integer index (no slices).

`Matrix` handles some of what `LEDMatrix` does now.  Note that this can directly use the free functions in `bibliopixel.matrix`, so I don't have to port any of that code!

Also, `Matrix` is not a `Strip` - it _contains_ a `Strip`, i.e. it writes onto a `Strip`.  This is by design.


rec:
No, my plan is to tease out your brilliant ideas that you have buried beneath a layer of unsurpassed ugliness. :-P

PR #90, opened 2017-01-03
Multimap code.

rec:
Note that the first commit fixes a regression :-/ from October!  Clearly no one caught this...


adammhaile:
What regression? That first commit doesn't seem to change anything old.
Otherwise +1

rec:
There is no variable `self.driver`.  The right name was `self.drivers`.

Had that code been executed, it would have failed!

adammhaile:
Ahh...oops. Gotcha

PR #94, opened 2017-01-17
Allow the use of shared memory.

rec:
This went a little too easily, so I'm suspicious, but it looks as if I'm testing it fairly thoroughly.  

Certainly it is clear that if you don't create a custom data_maker, everything is just the same!

This still doesn't let us go MP, but it's a big step in that direction.


adammhaile:
Looks great :) Merge at will.

PR #100, opened 2017-02-02
Setup.py cleanup and bug fixes

adammhaile:
Moved VERSION into bibliopixel dir
Updated VERSION string parsing methods
Added MANIFEST.in to have setup.py install VERSION to package
Updated setup.packages to actually install ALL sub packages
Disabled 2 line between classes flake8 warning
Added build dir to ignore


.gitignore
```diff
@@ -7,4 +8,4 @@ BiblioPixel.egg*
 *.pyc
 .cache
 .coverage
-htmlcov
\ No newline at end of file
+htmlcov
```
rec: Ah, p00p.  Sorry about that, I'll catch that in review next time.  Is there a way to quickly add a Travis rule here?

bibliopixel/__init__.py
```diff
@@ -2,9 +2,11 @@
...
 
+
 def _get_version():
     from os.path import abspath, dirname, join
-    filename = join(dirname(dirname(abspath(__file__))), 'VERSION')
+    filename = join(dirname(abspath(__file__)), 'VERSION')
     return open(filename).read().strip()
 
-__version__ = _get_version
+__version__ = _get_version()
```
rec: DOH!

setup.py
```diff
@@ -41,7 +41,12 @@ def run_tests(self):
                   coverage, fail_under))
             raise SystemExit(1)
 
-VERSION = open('VERSION').read().strip()
+def _get_version():
```
rec: Oh, I hate the near-duplication of code here.

Why do we have to move the VERSION into the Python directories anyway?  It isn't a .py file...?

tox.ini
```diff
@@ -1,4 +1,4 @@
 [flake8]
-ignore = F401,E501,E402,E401
+ignore = F401,E501,E402,E401, E305, E302, N812
```
rec: 1. You should keep the spacing consistent (nit pick).

2. Is there really a need to call the variable `VERSION`?  If we didn't do that, we could remove N812

3. This file suggests commenting each exception: http://flake8.pycqa.org/en/latest/user/configuration.html

4. Gee, E302 is silly!  :-)  We have to ignore that one.

5. We don't want to get rid of "E305 expected 2 blank lines after class or function definition, found 1", do we?  Why not just insert a blank line?

rec: Also, while you're there, might as well delete the next line, since it's commented out anyway.

.gitignore
adammhaile: weird... that wasn't what I changed. I added "build". Travis rule for what?

tox.ini
adammhaile: 2) Because that's what you called it :P I will change.
3) <raspberry noise> fine :P
4) Seriously. 
5) Wait... I thought that was E302? And the reason I was removing the check for 2 blank lines is that *you* weren't following it ;) My editor bitches at me when I do it so I was. But it showed a bunch of errors when I pulled the latest saying it needed 2 lines after a class.

.gitignore
rec: The last line didn't end with a carriage return.  If you look at the diff you see that little funny symbol that github uses to indicate its disapproval.

It's quite likely your editor - which is what btw? - just put that in automatically, as it should.

And of course, now I looked at this, of course emacs has something to do that too, so I'll turn that on after I press return.  :-D

adammhaile: Gotcha... yes, my editor always adds the newline on save. I use Atom. (I know I know... console editor or die. But they can pry Atom from my cold dead hands)

setup.py
rec: Oh, very interesting!  I just looked through my site-packages, and I see two sorts of things - directories like this, and single files.  So yes, you're totally right.

tox.ini
rec: I'm pretty sure the errors are:

    E302 expected 2 blank lines, found 0
    E305 expected 2 blank lines after class or function definition, found 1

Well, p00p!  I'm flaking out?

But from the moment you introduced me to flake8, I run it as part of my whole tests thing every single time.  So why wasn't it finding that missing line??

Under no circumstances allow me to slack off on flake8 styles.  It's very little effort to do it right.

rec:
:+1: Looks good to me if you make those tweaks.

tox.ini
```diff
@@ -1,4 +1,22 @@
 [flake8]
-ignore = F401,E501,E402,E401
-#exclude = test/*
+ignore =
+    # imported but unused
+    # __init__ often imports to make available
+    # but does not use
+    F401,
+    # Line too long
```
rec: National shame!!!

tox.ini
```diff
@@ -1,4 +1,22 @@
...
+    E501,
+    # module level import not at top of file
+    E402,
+    # multiple imports on one line
+    E401,
+    # class names should used CapWords
+    N801,
```
rec: When I'm bored I'll try to fix some of these.

PR #102, opened 2017-02-12
Add a test that loads all BP code, showing true coverage.

rec:
This commit came out while I was reviewing your large pull request so I decided to get it out fast.

It adds a single test that does nothing except load every piece of code possible.  This does two things - make sure it all at least loads and there aren't conflicts or even syntax errors - and makes our coverage over all our code and not just over "code that happens to be loaded as part of specific tests".

Our coverage now is around 40%...


adammhaile:
Nice

rec:
Travis failed!  Interesting... let me see!

rec:
So, sorry to get distracted on _this_, but I need this in order to show my patch for your big CL!

So, interestingly enough, I can reproduce this in a clean virtualenv!  If in BP.py I type:

     pip install -r test_requirements.txt

then it installs a bunch of stuff - but it does not install pyserial, even though that appears to be in requirements.txt.

Very interesting.  Glad I found this because some sucker somewhere will be having this issue too.  Now to figure out how to fix it.

adammhaile:
pyserial is in test_requirements.txt NOT requirements.txt

adammhaile:
err... scratch that, reverse it.

adammhaile:
And what do you mean to "show your patch"?

rec:
Oh - er - I have a patch for your change that removes the external SimpleWebSocketServer library you checked in and replaces it by a line in requirements.txt.

By "show my patch" I mean to prove that that patch really works - so that Travis would fail if my patch were broken.

But to show that is valid I need to have a test that exercises it.  So why not exercise all the code, at least by loading it?

Thus this.

adammhaile:
NO! That dev breaks it all the time. That's why I pulled it in. Not my
first time with that library. Sorry, but BDFL says it's gotta stay internal.

adammhaile:
I also have an incoming change that actually fixes some performance issues
with that library that the dev doesn't want. Was just waiting for this main
changeset to be done.

rec:
Heh, well, we can just comment that one line out too.  It's only a single line dependency, and it doesn't directly exercise anything I'm interesting.

It's an ittle bit of a pain because several drivers depend on that and they can't be smoke-tested either.  but, that's life.  :-D

However, I'm really interested as to why that one package doesn't load automatically.  Three dependencies, two load, no errors!

rec:
Well, it'll be a few lines in the test, but still, no problems.

adammhaile:
Comment out what? Now I'm confused about what part you are responding to.

rec:
You want to avoid needing a dependency on `pyserial` for Travis to pass, is what I'm reading.

I introduced this with this `all_test` thing.  I am going to remove it, but I'm also interested in finding out why Travis doesn't load pyserial when it's so clearly in the requirements.txt.

So far I have discovered that if I have a single line requirements.txt with just pyserial, it does work.  :-o

adammhaile:
Ok...weird. Wonder if pyserial just cannot be installed on their docker containers for some reason? Wait... you said it does with just that in the file. No idea then. Option would be to leave it in the requirements.txt with all the rest and then add a separate "pip install pyersial" line to .travis.yml

Anyways, mostly concerned with the SimpleWebSocketServer... no removing that! :) I'm just gonna put it back...

rec:
You shouldn't check in third party code just because the code part is one file and you can get away with it.

It should be loaded in requirements.txt like any other dependency!

Yes, I do do this sometimes in C++ but that's because there is no packaging system.  Python has a great packaging system and you're already using it.

Yes, I know this package isn't in pip, but you just add a specific commit ID to requirements.txt and it all works.

adammhaile:
Again... I have changes coming to this that I won't likely be able to get
the main dev to take. Consider this a fork for our own purposes. For
anything else I'd agree with you. But this stays.

rec:
You ain't giving me any tangible reason for doing it wrong - just fear, uncertainty and doubt.  :-D

However, I'm not going to hold it up of course, but I will be proposing a patch for it down the line.

You will lose your attachment to the dark side of the force and come to embrace correct packaging, young padawan.

adammhaile:
LOL... we run our own internal PyPi server at work because of the number of
times we've been bitten by packages out of our control or issues with being
unable to reach the real PyPi when we need to. Granted, we have hundreds
that need to be installed on dozens of testing runs each day. I'm not
opposed to it all, I'm just cautious of relinquishing control.

And the reason I'm giving you right now is that there will be updates to
this library that the official one won't have unless the main dev takes it.
If he takes it, by all means, let's pin it to that commit and move on. So,
yeah... that's uncertainty. I don't think he'll take them, but he might...
but for now there's nothing we can do but load my local version if I want
to use those changes. I was just holding off until this main update is in.
Or hell... would it make you feel better if I straight up fork the GitHub
repo and we load from that? But at that point, what's the point... we just
need to one file.

Side note, I'm straight up chaotic neutral with a slight lean towards evil.
I've embraced it. This is nothing new :P

Also, here's one of my issues with the dependencies in BiblioPixel... I
hate the number of dependencies that have to be installed that you may
*never* ever use! Pyserial is a great example. I'm all ears if you can
think of a good way to only have to install this stuff if and when you
actually need it. BiblioPixelAnimations is kind of this way. There's a ton
of stuff you would have to install if you want everything, but you may not
need them.

rec:
Good is better!

Basically, "evil" means "choosing to defect in the prisoner's dilemma".   But everything that humans have accomplished has involved choosing to cooperate - keeping mutually beneficial agreements even when defecting might involve better short term gains.

----

If you want to fork this guy's library, we should do that under the BiblioPixel username.  Perfectly legit.

If you make changes and want to have any hope of contributing them back upstream it's the only way to go!  Or even if you don't.

Suppose this guy releases some OK changes - you can rebase your fork against his and bring them in with no sweat.

There's no reason not to do it right.

---

I agree completely with your desire to avoid dependencies.

Indeed, I think we should remove _all_ of requirements.txt.  Seriously.

Those entries should go elsewhere - into "nice-to-have.txt" :-D or something - "optional.txt"?  Or just into test_requirements.txt

requirements.txt are the things that are _required_ for your library to be loaded.  I think _nothing_ in your requirements.txt is "required" for an installation of BP!

This is bad.  Suppose I write my own library using BP as a dependency. Unfortunately, this means that all this crap listed currently in dependencies.txt gets sucked in when I never use it.

---

I am going to move everything into test_requirements.txt for now.  I am going to see if I can get serial to load from that.  It's really weird that that doesn't work and I get no message or error.

Then I'll get back to you. DAMN, I wanted to write a coherent review before sending it and now you see the annoying opening and not the good parts at the end.  Sorry!!

adammhaile:
Haha... I know, I know... just messing with you. Just my D&D play style :P

Glad you agree on the non-required requirements! :) Maybe let's add some try/excepts around things that aren't strictly required that then print out "To use this module please: pip install <module>". I do this with Pyserial now. And then yes... remove requirements.txt (well, it should be in setup.py actually) entirely.

And yeah... I'm all for doing the web socket thing "right" if we can get there. I mainly didn't want to remove it _now_ as it's still in flux.

But yeah... I'm thinking a fork of that library is in order. For another day...

rec:
OK, I think you'll like this much better.  

requirements.txt goes entirely.  We test almost everything, and we have an expanded test_requirements.txt.

Travis passes, too.

adammhaile:
+1 works for me

PR #103, opened 2017-02-13
First half of "extracting threading".

rec:
So this is getting pretty long, and it's all refactorings, so let me get it out now and I can keep going on top of it.

This area in the code is delicate and I don't feel my battery of tests adequately tests it so I made a lot of very small, obvious moves.  Probably best to read one commit at a time as I suspect that each one is very uncontroversial.

The plan is to get all the code regarding threading into one file and I'm about two-thirds of the way there.  I tried hard to avoid anything other than that, but I tweaked the logic in one place to make that refactoring easier, and I found a tiny bug I fixed in passing.


bibliopixel/animation/base.py
```diff
@@ -44,7 +44,7 @@ def stopped(self):
 
     def _run(self, amt, fps, sleep, max_steps, until_complete,
              max_cycles, seconds):
-        self.preRun()
+        self.preRun(amt)
```
rec: Oops!

adammhaile: yeah... that was put there as an in case you wanted it. but no one uses it.

bibliopixel/animation/base.py
```diff
@@ -125,32 +97,15 @@ def run(self, amt=1, fps=None, sleep=None, max_steps=0, until_complete=False,
...
-                run()
-                log.debug('Thread Complete')
-
-            self._thread = threading.Thread(target=target, daemon=True)
-            self._thread.start()
-            if joinThread:
-                self._thread.join()
-        else:
-            run()
+        self._callback = callback
```
adammhaile: Just an FYI... nothing uses this callback that I'm aware of. It was kind of a good idea at the time but never got used. So this could be added to the list of things to prune if need be.

PR #124, opened 2017-03-04
Created bibliopixel system script

adammhaile:
Now, once installed, running `bibliopixel demo` will work from anywhere.
BTW, we could absolutely setup that demo script to loop over matrix, cube, circle, etc. SimPixel will automatically re-adjust with each change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,11 @@ from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 import sys
 
+if sys.version_info.major != 3:
+    print('BiblioPixel v3 is no longer supported on Python 2!')
+    print('If you need Python2 support please insteall BiblioPixel v2')
+    print('pip install "bibliopixel<3.0"')
+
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,6 +2,11 @@
 from setuptools.command.test import test as TestCommand
 import sys
 
+if sys.version_info.major != 3:
+    print('BiblioPixel v3 is no longer supported on Python 2!')
```
rec: Consider:  "Only losers use Python 2.  If you are a loser..."

j/k!

setup.py
```diff
@@ -73,5 +78,13 @@ def _get_version():
...
     },
-    include_package_data=True
+    include_package_data=True,
+    entry_points={
+        'console_scripts': [
+            'bibliopixel = bibliopixel.main.main:main'
+        ]
+    },
+    install_requires=[
+        'BiblioPixelAnimations'
```
rec: So it knows that this is your BPA because - ah, because you have installed it on pypi.  Got it.  

Looks good to me!

:+1:

PR #126, opened 2017-03-04
New `run` command for main.

rec:
You're going to like this!

You can now just say `bibliopixel run [project-file-name]` and it will run it.  PLUS, I now have good defaults for all parts of a Project, so `project-file-name` is optional.

Note that `project.py` refers to BiblioPixelAnimation - but _in text_.  This doesn't actually introduce a dependency yet on BPA, though it will try to import it if you don't fill in the animation yourself.

(So Travis continues to pass, even though it doesn't as far as I know bring in BPA...)


bibliopixel/main/main.py
```diff
@@ -3,7 +3,7 @@
 
 __all__ = ['main']
 
-COMMANDS = 'demo',  # 'device', 'network', 'run', 'settings'
+COMMANDS = 'demo', 'run',  # 'device', 'network', 'settings'
```
adammhaile: I realize this is valid python3 but i really hate this syntax. () around it would look cleaner in my opinion.

rec:
Fixed that and rebased!

PR #129, opened 2017-03-06
Whitespace-only changes.

rec:
Why just the unit tests?!

Something must be or have been weird here.  But what?

Is there a way to create a travis rule to prevent this?


adammhaile:
I'd imagine there's something we can add like with flake8 that enforces
whitespace. Will look into it.

PR #138, opened 2017-03-16
Depend on attrs Python library.

rec:
I don't think the commit to use attrs to describe devices is quite ready, so let me extract this nano-change so we can make a final decision on whether to use the attrs library (as you know, I vote "yes").

Documentation here: https://attrs.readthedocs.io


adammhaile:
Use attrs. I'm fine with it.

PR #139, opened 2017-03-18
Standardize time in seconds

### LINKED ISSUE #104: Standardize time in seconds.
BP uses two types of time - seconds and milliseconds - and there's a conversion factor of 1000 thrown around to keep things going in a dozen places in the code.

This is a trap.  We should just use Python standard times - seconds - everywhere.

rec:
* fix #104 
* other improvements


adammhaile:
+1 :)

PR #140, opened 2017-03-25
Simplify Project.

rec:
I "simplify" the class out of existence, preparatory to reading it all from the dotfiles.


rec:
(And this reason for this pull request seems a little obscure - I pulled it from a longer chain to make reviewing easier.  The exact point will be clearer, later...)

PR #147, opened 2017-03-30
Remove cruft from importer.py.

rec:
Nanochange left over from my attempt to do clever library importing.


rec:
I just pushed a substantive change after this!

This handles checking and reporting on both "whether you have a module installed" and "does it have a good version".

Next, I'm going to propose that we remove code like [this]( https://github.com/ManiacalLabs/BiblioPixel/blob/dev/bibliopixel/drivers/serial_driver.py#L11-L25) - and replace it with a simple `import` statement, relying on the new code in this pull request to catch the errors.

For 98% of the BP users, they shouldn't ever be writing a main program!  They should be just using a project, and if they want to write their own code, it should go either as a new class of driver, LED, or (most of the time) Animation.

And the remaining 2% will be able to figure out pretty easily what's going on if they can't import a module.

PR #150, opened 2017-04-13
Fixed demo list crash

adammhaile:
Renamed bloom to matrix
added more matrix demo animations
updated queue to actually work with BP3


rec:
Oops.  Sorry about that, but now we have animation/queue.py in the demos to test.

I don't suppose there's any way to add a unit test that also runs the animation queue?

Also - you could have left the original "bloom" demo and added a new "matrix" one - the more, the merrier!  But no biggie.

adammhaile:
I'll look at adding a unit test for queue... we're already doing it for a regular animation so that shouldn't be hard. But that's for another commit.
The reason I changed the demo name was:
A) There are 3 blooms... circle, matrix, and cube
B) I want to have demos for circle, matrix, and cube at the very least. Afterwards we can add others. Should probably also have a simple strip demo as well.

rec:
Sure, let's do it!! :+1:

PR #153, opened 2017-04-15
Fix integer issues in color math.

rec:
Took me too long to realize that the crash was in HEAD and not my code!

(Not sure how "demo" works for you...??)


adammhaile:
The demo doesn't scale the colors as it's always at full brightness. Which in my testing last night worked fine.
What do you mean not in your code?
+1 btw

rec:
Oops, not so clear on my part!  I meant I was preparing to make some other unrelated change and thought I'd caused this issue and it took me a bit to figure out that I wasn't.

I don't see how the demo could ever have worked though.  It's trying to index arrays with a non-integer!

rec:
Here's the command I executed that failed on dev/HEAD:

`PYTHONPATH=/development/BiblioPixelAnimations python /development/BiblioPixel/main.py demo`

adammhaile:
Not sure.   Automatic cast attempt? Well it's hardened against that now...

rec:
I'm on the brightness stuff then!

PR #155, opened 2017-04-15
Fix software brightness.

rec:
This seems to fix it.

I was wrong about `_buf` vs `_colors`.

`_colors` contains "a list of tuples" - "all the LEDs".
`_buf` belongs to just one driver.  It's a series of bytes representing the LEDs for that driver. 
`_packet` belongs to a single _serial_ driver.  Itis the final packet that is sent to the driver - basically `_buf` and any headers and trailing bytes.

`_buf` isn't a good name.  :-)


adammhaile:
Nope... But it's internal so let's not rename anything right now. +1

rec:
Dam' right.  "Proceed to the release!"

PR #160, opened 2017-04-17
Rename AnimationQueue to Sequence and fix #151.

### LINKED ISSUE #151: Projects should support queues.

rec:


bibliopixel/animation/sequence.py
```diff
@@ -37,15 +38,6 @@ def addAnim(self, anim, amt=1, fps=None, max_steps=0, until_complete=False,
     def preRun(self, amt=1):
         self.animIndex = -1
 
-    def run(self, amt=1, fps=None, sleep_time=0, max_steps=0, until_complete=False,
```
adammhaile: From my email: 
All that does is throw out everything passed in except for fps and until_complete and the run the rest with some hard coded values. This is required for keeping AnimationQueue API compatible with all other animations because users could pass other values in but nothing other than those really make sense....
The reason that fps is even used at all is that if you, for example, add all your animations with no specific FPS you can then specify FPS to the queue and all sub animations should run at that rate.
Note... We MUST keep AnimationQueue/Sequence API compatible with all other animations. It IS an animation like any other. Without that it cannot easily be run through things like PixelWeb.

bibliopixel/animation/sequence.py
```diff
@@ -61,4 +53,4 @@ def step(self, amt=1):
             run.update(threaded=False, join_thread=False)
 
             run['fps'] = run.get('fps') or self.fps
```
adammhaile: This is where the above comes in for overriding FPS.

rec:
My suggestion is that we check this in "as is" because "it works" and then make the fixes I suggested in the email thread...

adammhaile:
Works for me. +1

adammhaile:
Did you expect me not to? As long as that other stuff is implemented, it all works for me.

rec:
Sorry, I was answering a github prompt.  It asked me if I wanted to dismiss your review that asked for changes when checking in, I said yes, and it asked why!

adammhaile:
Oh...lol. Gotcha. Have never seen that.

rec:
First time I ever saw it either!!! :-o

PR #165, opened 2017-04-20
Fixes mainly for serial driver

adammhaile:
Added important imports to __init__ files
Fixed list_ports import
check for list_ports result being non-list on some platforms
removed too early set_brightness

Meant to submit this before we ran into the whole brightness thing... DriverSerial was totally non-functional.

### DIFF
--- a/bibliopixel/drivers/__init__.py
+++ b/bibliopixel/drivers/__init__.py
@@ -1,0 +1,1 @@
+from . driver_base import ChannelOrder
--- a/bibliopixel/drivers/driver_base.py
+++ b/bibliopixel/drivers/driver_base.py
@@ -47,7 +47,6 @@ class DriverBase(object):
         self.lastUpdate = 0
         self.brightness_lock = threading.Lock()
         self._brightness = 255
-        self.set_brightness(255)
 
     def set_layout(self, layout):
         pass
--- a/bibliopixel/drivers/serial/__init__.py
+++ b/bibliopixel/drivers/serial/__init__.py
@@ -1,0 +1,2 @@
+from . driver import DriverSerial
+from . codes import LEDTYPE
--- a/bibliopixel/drivers/serial/devices.py
+++ b/bibliopixel/drivers/serial/devices.py
@@ -4,7 +4,7 @@ from ... return_codes import RETURN_CODES, print_error
 from ... import log, util
 
 try:
-    import serial, serial.tools
+    import serial, serial.tools.list_ports
 
 except ImportError as e:
     error = "Please install pyserial 2.7+! pip install pyserial"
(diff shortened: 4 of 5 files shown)


bibliopixel/drivers/driver_base.py
```diff
@@ -47,7 +47,6 @@ def __init__(self, num=0, width=0, height=0, c_order=ChannelOrder.RGB,
         self.lastUpdate = 0
         self.brightness_lock = threading.Lock()
         self._brightness = 255
-        self.set_brightness(255)
```
adammhaile: Since this is run before the reset of each driver's __init__() code executes it was crashing due to DriverSerial not yet being connected to any device and since it uses set_device_brightness instead. But this wasn't really necessary anyways.

bibliopixel/drivers/__init__.py
```diff
@@ -0,0 +1 @@
+from . driver_base import ChannelOrder
```
adammhaile: ChannelOrder had to be imported as bibliopixel.drivers.driver_base.ChannelOrder when using it in a script which is no good.

bibliopixel/drivers/serial/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . driver import DriverSerial
```
adammhaile: Same issue as ChannelOrder import, bringing this back up to a reasonable place.

bibliopixel/drivers/serial/devices.py
```diff
@@ -33,7 +33,7 @@ def find_serial_devices(self):
             port = ports[0]
             id = self.get_device_id(port, self.baudrate)
             ver = self._get_device_version(port, self.baudrate)
-            if len(ports) > 1:
+            if isinstance(ports, list) and len(ports) > 1:
```
adammhaile: maybe this was fine on Mac but on Linux/Windows `ports` is not always a list (or it was never tested on Mac either).

bibliopixel/drivers/serial/driver.py
```diff
@@ -13,7 +13,7 @@ class DriverSerial(DriverBase):
     def __init__(self, type, num, dev="",
                  c_order=ChannelOrder.RGB, SPISpeed=2,
                  gamma=None, restart_timeout=3,
-                 deviceID=None, hardwareID="1D50:60AB",
+                 device_id=None, hardwareID="1D50:60AB",
```
adammhaile: renaming all these to be consistent... this failed because it was deviceID but you later use device_id, which didn't exist.

bibliopixel/drivers/serial/devices.py
rec: `ports` might be a tuple, though.  More Pythonic is to do something like this:

    if getattr(ports, '__len__', lambda: 0)():

or catch the exception.

adammhaile:
No worries :) I know *nothing* about developing too fast or anything :P
I updated to use your getattr suggestion. If you are cool with that I'll rebase this in.

rec:
It looks great!  Do it!

PR #167, opened 2017-04-22
fixup! Allow aliases for common types in projects.

rec:


rec:
Checks failed!  :-o  And it's not a "serial dependency" issue.  Let me see...

rec:
Fixed.

adammhaile:
+1

PR #170, opened 2017-04-23
Big animation refactoring (again).

rec:
Apparently you can't switch base branches after starting a pull request, even if they are the same commit ID. Ah, well!


adammhaile:
Oops. +1 again

PR #176, opened 2017-04-26
Only report frame overruns once.

rec:
Interestingly enough, I couldn't reproduce this issue today, which has happened every day up until now running `bp demo matrix_project`.

(Don't worry, I tested it by temporarily adding a `0` to the end of all the framerates.)


adammhaile:
+1

PR #177, opened 2017-04-28
Rename duplicate base.py files

rec:
My concentration is still somewhat shot so doing a simple chore to warm up, even though it doesn't really advance us to 3.0.

I made a bad decision having these two `base.py` files in `led/` and `animation/` so I gave them better names.


adammhaile:
+1

Issue #133, opened 2017-03-08
Many more demos.


rec:
We have the ability to add a lot of existing slick demos from BPA to the main demo program - we should do so because it's impressive and costs us little.

rec:
I'm unassigning myself here.  :-D

The existing "circle" demo doesn't work.  Indeed, I had likely never tried it because it was referring to a non-existent class in BPA and dies before it starts (see upcoming pull request to fix).

I fixed the gross error, but now it runs and does nothing.  Will defer to you @adammhaile on this one!

adammhaile:
Yeah... I need to make a BPA3 or something. Too much is breaking there are
no longer compatible with both. Will fix.

rec:
We can launch 3.0 with the demos we have.

adammhaile:
Done with 3.0.2

PR #178, opened 2017-05-01
Make default demo random, and fix non-existent symbol.

rec:
* The circle demo still doesn't work, but at least it doesn't crash:
  see #133.

* If no demo name is specified, the list of possible demos is printed and a
  random demo is selected - to encourage discovery of the features by the lazy.


bibliopixel/main/demo.py
```diff
@@ -143,8 +143,8 @@ def circle(args):
     driver = DriverSimPixel(sum(pixels_per), layout=points)
     led = LEDCircle(driver, rings=rings, maxAngleDiff=0)
 
-    from BiblioPixelAnimations.circle.bloom import CircleBloom
```
adammhaile: Dude.... Pull the latest BPA! This exists: https://github.com/ManiacalLabs/BiblioPixelAnimations/blob/master/BiblioPixelAnimations/circle/bloom.py#L5

rec:
OOPS!  fixing.

TBH, I had never run that before I wrote the randomizer, so I assumed there had been some sort of mistake.

--

So - how can I automate this process so this error is impossible going forward?  

1. We now have an effective dependency from BP to BPA so breaking changes to BPA are breaking changes to BP, but there is no announcement/version/branching.  (3.x issue but just so it's discussed...)

2. `bibliopixel demo` isn't covered by any sort of automatic testing, not even smoke testing (which would have caught this).

Indeed, bibliopixel/main has pretty poor coverage of any type.

3.x issue #179 covers both of these.

Now to fix the issues!

adammhaile:
Personally, I think smoke testing main will help. BPA is now pip installable so we can just pull it in on travis and test. Would've caught it here. And smoke testing main demo doesn't actually need to run SimPixel as that's just a fire and forget.

rec:
All right, fixed this and filed that.  :-D

PR #180, opened 2017-05-01
SettingsFile becomes PresetLibrary.

rec:
As I was writing the [documentation](https://github.com/ManiacalLabs/BiblioPixel/compare/dev...rec:help) (still in progress) it became clear that "settings" wasn't the best name.

I decided on a "library" of "presets" and changed the code to correspond.


adammhaile:
+1

PR #194, opened 2017-05-22
path_fixer sets and restores sys.path.

rec:
We don't actually use this code anywhere;  I wrote it as part of something else and then didn't use it.  It isn't so critical that we check it in, but I didn't want to just throw it out...


rec:
(Last in the series of "clean my branches" reviews.)

PR #196, opened 2017-05-24
Load projects from URLs.

rec:
I swore I'd do no new development but, well, I ended up having downtime waiting for something and just started to sketch.

Suspect you will like it, however - everything else flows through, but now you can specify a URL instead of a file path, so you can directly run a project file you checked into github somewhere.

There's one tiny tweak.  You can still enter "raw JSON" for the `bibliopixel run` command - but now you have to use the `-j` flag to do so.  Before the program "guessed" as to whether you were using JSON or not, but I think now there are more options, explicit is better than implicit.


adammhaile:
I should scold you for not finishing Wiki changes and instead doing this... but Ooooo shiny!
Since now you could do:
bibliopixel run https://raw.githubusercontent.com/adammhaile/BiblioPixelProjects/master/cube.json
Which is f*&$ing awesome. 

+1

rec:
You can even use `bibliopixel run https://github.com/adammhaile/BiblioPixelProjects/master/blob/cube.json` - in other words, the "naïve" URL that you get from github by defailt.

rec:
The advantage of your not paying me is that you can't stop me from doing fun things!  :-D

adammhaile:
What?! How does that work? Oh... haha, you put in special handling for GitHub. Missed that on first look. Very good Sir :) Good idea since we are all gonna do that :P

rec:
I intend to put in gitlab handling too - I think it might be identical logic!

adammhaile:
Nice. Probably so. They try to be a pretty close clone anyways.

PR #197, opened 2017-05-24
Now paths can refer to git repos.

rec:
Well, you shouldn't leave me alone in a room with an editor.

Yes, I know this is new material, but you are really going to like this - completely backwards compatible, and you can load in modules _right from github_ (or probably gitlab) without having to install them!

    bibliopixel --path=//git/github.com/ManiacalLabs/BiblioPixelAnimations run ...    

or for more complex needs:

    bibliopixel --path=//git/github.com/ManiacalLabs/BiblioPixelAnimations?branch=dev&commit=0b29a0b970ec1e662af34b74360f2fb31c70ce11 run ...


rec:
And it should work exactly the same from an entry in your project.json file, though I admit I haven't tested it.  :-)

rec:
Ah, one more comment.  This probably won't work on Windows unless you're on Cygwin or something.  It relies on the subprocess module calling some executable called `git` with certain command line flags to succeed.

Unfortunately there are no working Python 3 git/python interfaces that I saw.  We can figure out how to do it for Windows "at some point".

adammhaile:
Dude: https://pypi.python.org/pypi/GitPython/
I use it at work all the time. Py2 and Py3 support. Works great.
Though I admit you still have to have git installed. And actually that should probably work fine on Windows as long as you have it installed. I will test soon. You do know Windows devs use Git, right? :P Actually MS has gone full git for Windows itself.

Anyways, that's freaking awesome and super great for loading project files from URL because then it will also install the packages. 
I think the only other thing that would make it super awesomesauce is to allow pip installs... especially since BPA is pip installable.

+1 otherwise.

rec:
I started to sketch out a solution for pip this morning.  But I decided to hold off on this for now.

I wasn't entirely comfortable with having a _permanent_ side effect of simply running a document - one that might screw with other programs of yours.

Loading a commit into a directory and then referring to it in `sys.path` only affects the current job;  but loading a library into your pip environment or virtualenv affects all jobs that run in it thereafter.

Perhaps the program needs to `pip` at some point, but I'm not sure exactly how to do it.

---

I'm thinking to commit this as-is, and then replace the raw execute calls with GitPython - or actually, drop it in as an alternate transport layer, so to speak... :-)

rec:
I extracted out the git specific code into a module so that we can easily replace it with GitPython but I didn't manage to figure out how to check out a complete fresh module in GitPython - all the examples start from existing repos!

Mostly out of time on this for the next few days, considering checking it in in this known good state, but before I add GitPython.

rec:
OK, OK - I pushed another module that optionally uses GitPython.  Couldn't leave well-enough alone.

adammhaile:
LOL... are you done now? +1 Push it!

PR #204, opened 2017-06-01
Fixed cube demo for new sequence type

adammhaile:
new sequence changes broke the cube demo. Fixed it.

### DIFF
--- a/bibliopixel/main/demo.py
+++ b/bibliopixel/main/demo.py
@@ -107,7 +107,7 @@ CUBE_PROJECT = {
         'typename': 'sequence',
         'animations': [
             {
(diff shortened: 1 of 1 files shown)


rec:
:+1:

This is a triple failure on our part!  :-)

1. I wrote this wrong.
2. You missed it in review.
3. The code is not automatically exercised and hard to manually exercise in its entirety.

One of the reasons that #169 is so post-launch critical is that right now, "running each demo for two seconds" is impossible to achieve.

adammhaile:
Yup... agreed. We need a way of automatically running the demos. Definitely something for 3.1 :)

PR #207, opened 2017-06-06
Now config files must be JSON.

rec:
Part of my "delete things I can't be bothered to document" push.  :-D

Previously, config files could be in JSON _or_ in "configfile" format.  I don't want to document this, and we always do everything in JSON, so let's delete the unused code.


adammhaile:
Forgot you even had that other format. Yes, +1
.. json rules :P

PR #229, opened 2017-06-18
Add back C-based WS281X driver

adammhaile:
Plus a few other fixes and tweaks.
SPI driver should default to File interface which has no dependencies


bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
+# Original code by msurguy: https://github.com/ManiacalLabs/BiblioPixel/issues/51#issuecomment-228662943
+from . channel_order import ChannelOrder
```
rec: Plop a blank line after the comment, pls.

bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
...
+from . driver_base import DriverBase
+from .. import log
+from .. import gamma
+try:
```
rec: and before the code...

bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
...
+    from neopixel import Adafruit_NeoPixel, Color as NeoColor
+except:
+    log.error("""PiWS281X Requires the rpi_ws281x C extension.
```
rec: Hard to read, consider extracting the comment as a separate `ERROR` constant ("don't mix code and data")

bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
...
+Install rpi_ws281x with below commands:
+git clone https://github.com/jgarff/rpi_ws281x.git
+cd rpi_ws281x
+sudo apt-get install python-dev swig scons
+sudo scons
+cd python
+sudo python3 setup.py build install
```
rec: That last line is dangerous if they are using virtualenvs, as they would likely install it for a version of Python that isn't theirs... but not quite sure how to proceed?

Consider having two lines, one if you are using virtualenv and one if not??

bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
...
+    """)
+    raise
+
+
+class PiWS281X(DriverBase):
```
rec: Why does the name start with `Pi`?

bibliopixel/drivers/PiWS281X.py
```diff
@@ -0,0 +1,52 @@
...
+    def _compute_packet(self):
+        self._render()
+        data = self._buf
+        self._packet = [tuple(data[(p * 3):(p * 3) + 3]) for p in range(len(data) // 3)]
+
+    def _send_packet(self):
+        for i in range(len(self._packet)):
+            c = NeoColor(self._packet[i][0], self._packet[i][1], self._packet[i][2])
+            self._strip.setPixelColor(i, c)
+
```
rec: The above three lines of code is not so Pythonic.  Consider:

    for i, p in enumerate(self._packet):
        self._strip.setPixelColor(i, NeoColor(*p))

bibliopixel/drivers/driver_base.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, num=0, width=0, height=0, c_order=ChannelOrder.RGB,
         self.lastUpdate = 0
         self.brightness_lock = threading.Lock()
         self._brightness = 255
+        self._waiting_brightness = None
```
rec: Whoa!  Well, that sux that that wasn't there!  :-o

bibliopixel/drivers/spi_driver_base.py
```diff
@@ -50,11 +50,12 @@ def __init__(self, **kwargs):
 
         try:
             from periphery import SPI
-            self._spi = SPI(self._dev, 0, self._spi_speed*1e6)
+            self._spi = SPI(self._dev, 0, self._spi_speed * 1e6)
         except ImportError:
```
rec: :-)

bibliopixel/drivers/spi_driver_base.py
```diff
@@ -50,11 +50,12 @@ def __init__(self, **kwargs):
...
             error(CANT_IMPORT_PERIPHERY_ERROR)
 
-        log.info('periphery spi dev {:s} speed @ {:.2f} MHz'.format(self._dev, self._spi.max_speed/1e6))
+        log.info('periphery spi dev {:s} speed @ {:.2f} MHz'.format(
```
rec: Consider extracting this duplicate code (also on 100-101) into a method...

bibliopixel/drivers/PiWS281X.py
adammhaile: Blank line after try? Never.

bibliopixel/drivers/PiWS281X.py
adammhaile: Because it ONLY runs on the Raspberry Pi.

bibliopixel/drivers/PiWS281X.py
rec: No, no - before the `try`, after the imports.

bibliopixel/drivers/PiWS281X.py
rec: Hah! well, then.  Perhaps a tiny comment?

bibliopixel/drivers/PiWS281X.py
adammhaile: fair enough... I didn't write it originally :) that's my defense anyways.

bibliopixel/drivers/driver_base.py
adammhaile: Yeah... I really don't know why it wasn't. It would throw an exception if you didn't set the brightness before running animations on some drivers. It's the one thing I hate about Python that you can use vars that were never initialized previously.

bibliopixel/drivers/spi_driver_base.py
adammhaile: I prob should've made this a separate commit... but I was there.

bibliopixel/drivers/spi_driver_base.py
adammhaile: I still plan on a separate commit where I rework the SPI stuff... for one, it's unusable in projects not because the ```interface``` param takes a class! That ain't gonna work :P So I have to make that an enumeration or something. For the record, I do the same thing all the time... it's easier, but just incompatible with projects... that commit is next after this.

bibliopixel/drivers/PiWS281X.py
adammhaile: Yup, and I'll make a Wiki entry for it.

adammhaile:
Rebased in requested changes... to hide my shame.

rec:
:+1: !

Rebasing is close to godliness.  :-D

adammhaile:
LOL :)

PR #238, opened 2017-06-23
Variety of fixes to compatibility with projects

adammhaile:
Found while working through some wiki fixes. Relatively simple fixes but major consequences to project support.

Added **kwds to all drivers. Fail with maker keyerror otherwise
Removed ImageSequence. Deprecated and broken
Added string c_order handling to all drivers
Allowed setting matrix map options from project


adammhaile:
Self review! Don't kill me @rec - stuff was broke.

PR #248, opened 2017-06-27
APA102 Cleanup and Fixes

adammhaile:
- Fixed APA102 bug for FILE mode (bytearray fail)
- Combined APA102/SK9822 driver
- Added chipset level brightness control
- Moved to full list slicing for greater speed (500% faster!)


adammhaile:
Quick demo: https://goo.gl/photos/v9UJgNkJBg2tkz4e6
The square display on the left is the ULTiM8x8 (APA102C) and the long strip is a 144/m SK9822 both running off the same data signal and both scaled on the chipset to 12.5% (32). Both chipsets actually only do 5 bit brightness but set_device_brightness() handles the conversion (`val >> 3`). Even still, the on chip scaling looks WAY better than scaled in BP.

rec:
Probably at least part of that is that you still have the lousy integer arithmetic in BP still. :-D :-D

On to this review now!

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
 from . base import SPIBase
 
 
 class APA102(SPIBase):
-    """Driver for APA102 based LED strips on devices like the Raspberry Pi and BeagleBone"""
+    """Driver for APA102/SK9822 based LED strips on devices like the Raspberry Pi and BeagleBone"""
 
     def __init__(self, num, gamma=gamma.APA102, **kwargs):
         super().__init__(num, gamma=gamma, **kwargs)
+        maker = kwargs.get('maker', data_maker.MAKER)
```
rec: Perhaps rather DriverBase should store the `maker` as a member variable?  I can imagine that we'll do this again...

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
-        # APA102 requires latch bytes at the end
-        self._latch_bytes = [0xFF, 0, 0, 0] * ((num // 64) + 1)
+        # APA102/SK9822 requires latch bytes at the end)
+        # Many thanks to this article for combined APA102/SK9822 protocol
+        # https://cpldcpu.com/2016/12/13/sk9822-a-clone-of-the-apa102/
+        self._start_frame = 4  # start frame is [0, 0, 0, 0]
+        self._pixel_bytes = self.bufByteCount() + self.numLEDs  # 1 extra for brightness
+        self._pixel_stop = self._start_frame + self._pixel_bytes
+        self._reset_frame = 4  # for SK9822 [0, 0, 0, 0]
+        self._end_frame = ((num // 2) + 1)
```
rec: Quibble - lose the outer parens.

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
     def __init__(self, num, gamma=gamma.APA102, **kwargs):
         super().__init__(num, gamma=gamma, **kwargs)
+        maker = kwargs.get('maker', data_maker.MAKER)
 
-        # APA102 requires latch bytes at the end
-        self._latch_bytes = [0xFF, 0, 0, 0] * ((num // 64) + 1)
+        # APA102/SK9822 requires latch bytes at the end)
+        # Many thanks to this article for combined APA102/SK9822 protocol
+        # https://cpldcpu.com/2016/12/13/sk9822-a-clone-of-the-apa102/
+        self._start_frame = 4  # start frame is [0, 0, 0, 0]
```
rec: This is the same `4` that you see in line 20 - consider making it a class variable, like `FRAME_SIZE`.  Again, just a quibble.

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
+        self._pixel_bytes = self.bufByteCount() + self.numLEDs  # 1 extra for brightness
```
rec: The comment says 1 extra, but I see no +1?

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
+    def set_device_brightness(self, val):
+        """
+        APA102 & SK9822 support on chip brightness control allowing greater color depth.
+        APA102 superimposes a 440Hz PWM on the 19kHz base PWM to control brightness.
+        SK9822 uses a base 4.7kHz PWM but controls brightness with a variable current source.
+        Because of this SK9822 will have much less flicker at lower levels.
+        Either way, this option is better and faster than scaling in BiblioPixel
+        """
+        self._chipset_brightness = (val >> 3)  # bitshift to scale from 8 bit to 5
+        self._brightness_list = [0xE0 + self._chipset_brightness] * self.numLEDs
```
rec: What's the magic `0xE0` about?

bibliopixel/drivers/SPI/APA102.py
```diff
@@ -1,24 +1,43 @@
...
         self._render()
-        # TODO: Do all this surgery in place instead
-        newBuf = [0xFF] * (self.bufByteCount() + self.numLEDs)
-        newBuf[1::4] = self._buf[0::3]
-        newBuf[2::4] = self._buf[1::3]
-        newBuf[3::4] = self._buf[2::3]
-        newBuf = [0, 0, 0, 0] + newBuf
-        newBuf.extend(self._latch_bytes)
-        self._packet = newBuf
+        self._packet[self._start_frame + 1:self._pixel_stop:4] = self._buf[0::3]
```
rec: This code is both beautiful and ghastly at the same time!  More beautiful than ghastly when it comes down to it - sometimes you gotta do surgery.  And 500% faster!  I'm not really surprised, to be honest - all those memory allocations aren't free.

bibliopixel/drivers/SPI/APA102.py
adammhaile: Good point... will update.

bibliopixel/drivers/SPI/APA102.py
adammhaile: on APA102 per the spec they are actually different for start and end. On SK9822 they are also not really the same either but the result of making a protocol that tricks both into work is that they are the same. Which is why I still kept them different.

bibliopixel/drivers/SPI/APA102.py
adammhaile: Ohh.... ha, yeah. That's kind of a cheat... bufByteCount() is (numLEDs * 3) - There was plans to maybe support layouts with different numbers of colors but that really never happened - But the APA102 has a 4 byte frame, so: (numLEDs * 3) + numLEDs = numLEDs * 4.
That's old code. Was trying to respect the bufByteCount thing. Which doesn't really wake a lot of sense. Will change.

bibliopixel/drivers/SPI/APA102.py
adammhaile: Read this: https://cpldcpu.com/2016/12/13/sk9822-a-clone-of-the-apa102/
Basically, the first byte of the 4 byte data frame is brightness, but they only support 5-bit brightness control. Every other driver for APA102 you will see just sets that to 0xFF (full brightness) and then handles brightness in software (yes, we are the first to FULLY support this chipset!). So, per the protocol spec, 0xE0 is 0% brightness and then you can add up to 31 to that to get full brightness. This is why I just take the 8-bit value we normally use and do the `>> 3` to it, to scale it to 5-bits.

bibliopixel/drivers/SPI/APA102.py
adammhaile: Yeah... I was only going to fix the FILE issue and maybe add brightness. But I realized that to add brightness it was just easier to do the surgery (which I just learned you are the only one that calls it that... it's technically called slicing). But yeah, for 1000 pixels it used to take 95ms to update, and not it takes ~15ms.

adammhaile:
The FILE fix was actually the `maker` thing. I rebased down to one commit so you ended up not seeing the original single fix. Basically because APA102 was doing it's own thing with the buffer, it was returning a list and not a bytearray (which maker creates). But the FILE interface requires a bytearray. So this fixes that.

adammhaile:
Anyways... quibbles fixed. Should be good to go now. Will make this 3.0.5 - Which, btw, I vote that when we do version number updates, we leave that as a single commit and no PR necessary. Or at least no review necessary. I just don't want to also have to remember to change VERSION in the initial PR (I forgot this time). And it's a little cleaner to be able to decide if a PR is a new version to release after the fact.

adammhaile:
BTW... I will combine LEDTYPE values in a later release but as this fixes an actual bug we need to release this as 3.0.5 now not later as it could impact users.

rec:
:+1:

rec:
And I 100% agree with the single commit/no review necessary for version numbers.

PR #249, opened 2017-06-27
Special values in projects (fix #242).

### LINKED ISSUE #242: projects.py level global pre-processing of special values
Thanks to @haooken for pointing this out.

There are several special types of values that are settable inside a project by using a more raw representation but not by a nice, human friendly, value. We want projects to be human friendly.

For example, colors. Right now you have to enter [255, 0, 0] instead of colors.Red.

All of these special types internally are part of an enum or class so we can key off that.

Options are:

colors
LEDTYPE (this is a little weird becacuse there's two... serial and SPI)
Rotation
gamma

Probably more. Can't find them right now.
What I recommend is we register those types then look for the type followed by a period. If so, we try to resolve and replace that value inside the dictionary. 

Yes, this requires blindly walking the whole dictionary. But that's not really too hard.

This will make projects much cleaner to use and read.

rec:
All right!  Worked out quite well.  I sort of overengineered "parsing time", but I still like the code.

You can enter the values in human-readable form - like [this](https://github.com/rec/BiblioPixel/blob/special-values/test/project_test.py#L30-L37).

There's a very orderly way of doing the parsing, with modules containing single-dispatch functions - [like this](https://github.com/rec/BiblioPixel/blob/special-values/bibliopixel/project/types/gamma.py) to parse gamma arguments.

I realize now that I'm missing rotations, but I'm going to create an issue for this and assign it to @adammhaile just so that the knowledge of how to make these gets spread around...


bibliopixel/project/project.py
```diff
@@ -18,7 +22,7 @@ def multi_drivers(device_ids, width, height, serpentine=False, **kwds):
 
         for id in device_ids:
             build.addRow(gen_matrix(width, height, serpentine=serpentine))
```
adammhaile: Oh yeah... what is the project file syntax for multiple drivers??

bibliopixel/project/types/defaults.py
```diff
@@ -0,0 +1,12 @@
+from . import channel_order, color, duration, gamma, ledtype
+
+FIELD_TYPES = {
+    'channel_order': channel_order,
```
adammhaile: No param is ever channel_order, it's always c_order

bibliopixel/project/types/duration.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+USAGE = """A time can be initialized with:
+
+* A number that is greater than or equal to zero, representing a time in
+  seconds.
+
+* A string:
+
+2 weeks, 1 day, 1 hours, 3 minutes, 45 seconds, 340.4ms....
```
adammhaile: This all for the 'seconds' param of run() on animations? Will it even support 340.4ms?

bibliopixel/project/types/ledtype.py
```diff
@@ -0,0 +1,16 @@
+import functools
+from ...drivers.serial import codes
+
+USAGE = """An LEDTYPE is represented by a string.
```
adammhaile: Ah yeah... I forgot in that last PR I wanted to consolidate this with SPI_LEDTYPE: https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/drivers/SPI/driver.py#L13
I can make them the same easily. But I'll leave that for another future PR.

bibliopixel/util/duration.py
```diff
@@ -0,0 +1,106 @@
+import re
+
+"""Convert a string duration to a number of seconds.
+"""
+
+SI_PREFIXES = (
+    (10 ** 24, 'yotta', 'Y'),
+    (10 ** 21, 'zetta', 'Z'),
```
adammhaile: REALLY? So we run an animation for 1Zs and it's still going at the heat death of the universe...
Also, the ones at the other end of the time scale. These are completely useless for any real world usage. I get the desire for completion but this is just kinda silly.

test/project_test.py
```diff
@@ -19,6 +21,30 @@
 }
 """
 
+PROJECT_TYPES = """
+{
+    "driver": {
+        "typename": "dummy",
+        "num": 12,
+
+        "channel_order": "brg",
```
adammhaile: Again, nothing actually has a channel_order param.

bibliopixel/project/project.py
rec: I filed #251 here.

bibliopixel/project/types/defaults.py
rec: Costs nothing to have this.  And as you know, I prefer more explicit names that don't need documentation...

adammhaile: Well, yes, but there's nothing that can actually use it :P I'm fine with leaving it. Maybe in the future we'll make either c_order or channel_order work.

bibliopixel/util/duration.py
rec: Well, I just pasted an existing table into here with a little editing.  :-D

Cost me nothing to do all of them, so why not?   I personally want yottasecond animations, and I run them very very fast, yo.

Easiest to do it all.

rec:
Pushed a new version w/o channel_order.

PR #255, opened 2017-06-28
Replace type with ledtype in a few places.

rec:


bibliopixel/drivers/serial/driver.py
```diff
@@ -21,19 +21,19 @@ def __init__(self, ledtype, num, dev="",
         self.devices = Devices(hardwareID, baudrate)
         self.serial = self.devices.serial
 
-        if spi_speed < 1 or spi_speed > 24 or not (type in SPIChipsets):
```
rec: This was wrong, but it didn't fail, because `type` is a built-in name!

bibliopixel/drivers/serial/driver.py
```diff
@@ -145,7 +146,7 @@ def _send_sync(self):
 class TeensySmartMatrix(Serial):
     def __init__(self, width, height, dev="", device_id=None,
                  hardwareID="16C0:0483"):
-        super().__init__(type=LEDTYPE.GENERIC, num=width * height,
```
rec: This would have failed if using a project.

bibliopixel/project/project.py
```diff
@@ -72,13 +72,8 @@ def _apply_defaults(project, defaults):
...
     ledtype = defaults.get('ledtype')
     if ledtype:
-        if 'layout' not in project:
-            project['layout'] = {}
-
-        if 'type' not in project['layout']:
-            project['layout']['type'] = ledtype
-            # The field in LAYOUT should probably be renamed "ledtype" to be
-            # consistent across the whole system.
+        if 'ledtype' not in project['driver']:
```
rec: This was just plain wrong.  Layouts don't have types or ledtypes!

bibliopixel/drivers/serial/driver.py
adammhaile: ack! nice catch

bibliopixel/drivers/serial/driver.py
adammhaile: ah, very true. Again, nice catch.

bibliopixel/project/project.py
adammhaile: LOL... I didn't do that ;)

rec:
Yes, I know very well who did that.  :-D

PR #265, opened 2017-07-04
Improved multiple driver projects (fix #264)

### LINKED ISSUE #264: Allow projects with multiple, heterogeneous drivers.
BP projects only support multiple drivers if the drivers differ only by device id - that means if things like width, height, or serpentine change are different, you're out of luck.

This also results in a hard-to-explain data format.

We should fix this, and then fix #263 with the new documentation.

rec:
I had to be backward compatible, but I wanted to avoid guessing the user's intentions from the data too much.  This is a compromise.

I'll edit the Wiki once this pull request is approved, but here's the sketch.

Before this change, a project had a required `driver` section.  Now `driver` is optional, and a project can also have an optional `drivers` section.

If `drivers` is not set, everything works exactly as before - `driver` is required, and that dictionary is used to construct a single driver.

if `drivers` is set then it is a list of dictionaries used to construct a list of drivers.

Each driver in this list _must_ have non-optional `width` and `height` entries.  That `width` and `height` are passed to the driver constructor _and_ used to add rows to the `MultiDriver`.

Additionally, there is an optional `serpentine` parameter for each driver that is used in adding rows to the `MultiDriver` but is _not_ passed to the driver constructor.

If both `driver` and `drivers` are set, then the dictionary `driver` is used as a default for each element of `drivers`.  This is extremely useful for cutting down duplication when there are multiple drivers which differ only in one or two entries (like `deviceID`).

Example of the new format:

    {    
        "driver": {
            "typename": "dummy",
            "multi": true,
            "width": 128,
            "height": 32,
            "num": 4096
        },

        "drivers": [
            {"device_ids": 10},
            {"device_ids": 11},
            {"device_ids": 12}
        ]
    }

Compare with [this](https://github.com/ManiacalLabs/BiblioPixel/blob/03325d854b42edcf051709ad8bb3829864c211c9/test/project_test.py#L48-L55).


rec:
I found a little issue.  It won't involve changing this file.  I'll push a fixup but you can review this while I do.

adammhaile:
Why is serpentine a special thing?
Can I still just specify `drivers` and do all individual entries even if there's duplication?

rec:
`serpentine` is a special thing because ["it was before"](https://github.com/ManiacalLabs/BiblioPixel/blob/03325d854b42edcf051709ad8bb3829864c211c9/bibliopixel/project/project.py#L16-L37) but looking at the code for multimaps for projects, this has always been wrong.   :-o

I'm going to see when I made this mistake and what was there before to lead me to do this.  But I'll also fix this in this review.

As to the second question, yes, you can specify `drivers` and not `driver` if you like, and have no defaults, even if there's duplication.

rec:
Ah, I was wrong - I misread the code.  

`serpentine` is perfectly legitimate - however there are [three more parameters](https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/layout/geometry/matrix.py#L48-L49), `offset`, `rotation` and `y_flip`.

Instead of adding three more "synthetic" values to the dictionary, I put them into a new entry called `matrix` which is just passed along as arguments to `gen_matrix`.

I also fixed up alias resolution so it works for `typename`s that are found in the drivers list.

rec:
All right!  Ready to go with tests.

adammhaile:
My concern with serpentine et al is that there is also cube and circle. How do we handle those?

adammhaile:
Wait, wait... why is this on drivers at all????? pixel layout is for layouts! Why is this on drivers? Serpentine is for Matrix and Cube, not drivers.

rec:
Well, I would again say, "This is how it always has been."  We're not actually doing anything new - we're doing the same thing with a different format input.

The way someone other than myself (cough) originally set this up in BP2.x, `serpentine` (and those three other parameters) appear in _two_ places.

They appear as arguments to `layout.Matrix` (originally `led.Matrix`), which goes on to call `gen_matrix`, and they appear as arguments to `gen_matrix` directly.

We call `gen_matrix` directly when laying out multidriver matrices under one `Matrix` in creating projects.

That code, iterating over multiple drivers and using `gen_matrix` to lay out, originally came from some other source than myself, perhaps demo code you had somewhere for the 12k?

rec:
As for cube, sphere and strip, we have no multidriver support for them yet.

adammhaile:
Ok, I get that... but let's move serpentine out of driver for projects, it *must* be in the layout section. In the code, it's in the layout, not the driver.

rec:
Here's the example to keep in mind:  two different driver setups with the same layout: 32 lights in an 8 x 4 matrix.

1. one 8x4 driver, wired serpentine.
2. two 4x4 drivers, each wired serpentine.

A few moments' thought should convince you that if the serpentine parameter is only on the layout, you cannot map the second driver layout correctly (and it's even worse for oddball installations like "1 serpentine, 1 not" which we should also support).

And if you copy the serpentine parameter from the layout to the driver(s), you get a completely botched layout.

I agree we need to handle this better, but this change can't do that.  It's just allowing us the same problem we had before with a different data input.

adammhaile:
When you do this in code, you generate the layout separate of the driver
and pass it all to layout. The layout figures out which  pixels go to which
driver based on how many pixels each driver is supposed to have.
So, yeah, I guess it makes sense to have the mapping stuff with the driver
they go with and then merge it all up. I just messes with my head since the
mapping has always been passed to the layout, not driver. I get where you
are going with this... but yeah, it's not a great solution. It needs good
docs at least.
+1... merge it.

rec:
Oi!  Don't blame me here!  :-D  This is all derivative of some chunk of code you sent me as an example of how to run a multidriver.  I really haven't changed any of the algorithms at all.  Indeed, I haven't really thought it through too much...

My hope is that when we pull out the things like rotation, serpentine and reflections as separate free functions that we can use as adaptors, this will all become clearer... but who knows?

rec:
By the way, this was very instructive for me at least on how this layout thingie worked.

adammhaile:
No blaming you... just thought it weird to have mapping in the driver
section, but yet, in this case it does make sense.

PR #275, opened 2017-07-08
Fix #261 again.

### LINKED ISSUE #261: Command line flags are position-dependent.
`bibliopixel  --verbose run --animation=matrix_check` works, but

    $ bibliopixel run --verbose --animation=matrix_check 
    usage: bibliopixel [-h] [--loglevel {debug,info,warning,error,critical}]
                       [--path PATH] [--verbose] [--version]
                       {all_pixel,clear_cache,demo,devices,run} ...
    bibliopixel: error: unrecognized arguments: --verbose

NOTE:  this is a known issue with how `argparse` works with separate commands, not a bug in our code.  However, I can see two ways of getting around this...

rec:


adammhaile:
Can you give me a rundown of how, if at all, things work differently now? I think I get it from the code, but want to make sure we are on the same page.

rec:
Sure!

Before what I did was to attempt to gather up flags after the command that should be before the command and put them before the command.  argsparse wants top-level, universal flags to be before the command and flags for a specific command to be after the command.

(Really, `argparse`, that's a shitty feature.  Why can't I just turn that off? Why am I wasting my time on this??)

Copying and rewriting `sys.argv` into some better form is obviously stupid and I blame you for approving this. . :-P

And even worse now given that those flags either include their own value, or have it as the next command line argument.

So no more rewriting the arguments.  Much simpler code.

Now I instead add _all_ the common flags to _each_ command separately!  That means that flags can _only_ be after the command - very consistent.

`--version` is still a special case because we want to allow `bibliopixel --version` and have it just print the version and stop.

adammhaile:
Cool +1

adammhaile:
Rebasing in myself because I need to mess with main.py

PR #278, opened 2017-07-09
Correct pip name of noise module.

rec:


rec:
To be honest, we don't actually have to name modules whose pip install name is the same as their name.  But I like doing this.  Later we might distinguish between "unknown modules" and "ones we know about".

PR #287, opened 2017-07-14
bibliopixel upgrade (fixes #279)

### LINKED ISSUE #279: `--upgrade-depends` flag to `bibliopixel`
See [this thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/bibliopixel-dev/zobt4Vl6QmE/bAuouq-5AAAJ).

rec:


bibliopixel/main/update.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+import os, sys, subprocess
+
+
+def run(args, settings):
+    root = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+    code = 0
+
+    with open(os.path.join(root, 'requirements.txt')) as f:
```
adammhaile: Just a note that this only works because we include requirements.txt in MANIFEST.in
But please test this by running `python setup.py sdist`, then pip installing the generated tar.gz file in a venv. Only way to make sure this will actually work when uploaded to pypi.

rec: Wow, good to know.

Very clear instructions - tried it, and it worked.

adammhaile:
Awesome. +1 then :)
Just been bit by this a bunch... my `remote` branch actually has fixes for
some setup.py stuff on Windows. It's a fickle beast.

rec:
We should try to extract that code out... somehow, so that we can make sure that I don't foolishly screw things up.

PR #293, opened 2017-07-18
Allow projects to use numpy arrays.

rec:
"Sooo, sooo difficult!  Only in version 4!" -someone.  :-D

Turns out it's just another flag to the data_maker.

Of course, this flag is turned off by default so no behavior changes, but you can turn on numpy as part of your maker flags in your project, or with the command line `--numpy`/`-n` flag.


adammhaile:
LOL... Still getting used to the maker thing. Last time we discussed it numpy was going to be a pain.
Granted, these are still Int aren't they? Or at least they need to get converted to int values at some point as the LEDs don't take float.

rec:
No, if you turn on numpy then by default it's float, though you can override that in the project.

Everything does at the end get converted to int for the drivers.  

Indeed, I suspect we round off almost all our intermediate calculations, so for the moment you might not expect to see any different results.

However, now we have the possibility that a `ColorList` might be a `numpy.array` we can get rid of rounding where appropriate later on and start to get better gamma(*) and fades.

(If you recall, once you apply gamma=2.5, there are over a dozen output levels on the LED that are simply never even hit, reducing the output gamut by about one bit.)

Issue #298, opened 2017-07-20
"remote" should be an animation.


rec:
`_make_remote` just returns an animation.  From reading this code several times, I see no reason why we shouldn't just have a new sort of animation, a `RemoteAnimation`, which contains other animations.

rec:
The reason this comes up is if we don't do this, we're going to need a new section in the project for each different type of externally-controlled animation - and that might be a lot.

adammhaile:
Well it returns something that has start and cleanup functions. But yeah...
I get what you mean. Should be easy... But Id have to look at how that
might complicate the config...it will certainly me less simple but prob not
too bad. I'll see if I can make that work before 3.1 release. Shouldn't be
too hard...

rec:
:+1:

rec:
I looked at this some more.

It's not so much code I think.

You derive `Remote` from `Sequence` and simply construct everything as if it were a `Sequence` - you only override `step`.

Let me do an introductory tiny change to threading to make this easy for you.  Moment pls.

rec:
OK, no change coming.

The one issue is ensuring that child animations of a remote animation are forced to be threaded.

There's no slick way to do this right now.  The easiest way is to override `Animation.set_runner` in `Remote` and after calling `super().set_runner`, "manually" setting this value on each child animation.

adammhaile:
No! Remote IS NOT a Sequence.
Remote animations are individual.

rec:
Then derive from the new `bp.animations.collection.Collection` class.  :-D

adammhaile:
Why? That just complicates what I have.

rec:
Why does it make it more complex?

All that code you have constructing subanimations for remotes shouldn't be there in that code body.

We are already construction subanimations - for Sequences.  We should be doing it the same way every time, not having two different ways of doing it depending on the class.

rec:
Instead of yakking about it, I can just do it.  :-D  Moment pls.

adammhaile:
See comment on your PR.

rec:
This is fixed, yes?

adammhaile:
Yup, of course :) One of these days I'm going to make a project file with a
remote that starts a remote that starts a remote that starts a remote (all
on diff ports of course).

PR #299, opened 2017-07-20
Extract out a base class for collections of animations.

rec:
See #298.


adammhaile:
What actually happens with run() is called? How do I tell it to switch to an animation?
With this I cannot call cleanup on the running animation (this is absolutely required) when I switch to another animation.

adammhaile:
+1

PR #301, opened 2017-07-22
Suppress Travis compilation for "experimental" branches.

rec:
Feel free to reject this!

I do rapid development in experimental branches where I don't want travis going off all the time, so I set up a rule to suppress compilation just for branches whose names start with `x-`, `y-` or `z-`, which I deem "experimental".  :-D

(I know I can also do this by putting a tag in the commit ID but there are often dozens of these little commits which I create and destroy very fast...)

This is also a chance to review the .travis.yml file for me...

0. How do we test changes to this file?
1. Why don't we also build for 3.6?
2. What does it mean that the deploy version for Python is 3.5?


adammhaile:
Why x-, y-, or z-?

0. Easy... enable travis on your fork. Then it will run on all this stuff.
1. No reason, probably should build for 3.6
2. I explained this before. The deploy version is inconsequential as it's an sdist, but there's a pypi bug in 3.4 so I deploy from 3.5. It will not deploy unless all succeed however.

rec:
Travis is enabled on my fork.  So... I need to create a branch with a name x- but this .travis.yml and see if it gets built...

`x-` is short for experimental, but I found myself using `z-` too.  As I said, this is pretty small beanz and if you hated it, I'd dump it - it'd just save me a bit of spammy clutter.

adammhaile:
naw, you can keep it in there. Works for me.

rec:
The tests did work.  I created two broken branches with different commit IDs, one for a branch starting with `t-` and one starting with `x-` and then waited for Travis.  Travis complained about the `t-` and ignored the `x-` branch.

PR #304, opened 2017-07-22
fixup! Add a unit test for sequences and fix some issues.

rec:
Oh, how embarrassing.  

With your permission, I'm going to actually fold this into dev/HEAD so it will be as if this mistake never happened - not to obscure blame, but rather to not have a bad commit ID in the list (even if we aren't currently testing for that problem).


rec:
Urg, on the travis breakage, moment pls.

rec:
Fixed.

rec:
So shall I rebase this over dev, then??

If not, I'll give the commit a better name.

adammhaile:
I'd rather it just PR in like normal... Then I don't have to force pull or
anything.

rec:
Wilco!  I have a guest here, this won't happen for a bit...

adammhaile:
I need to get on with remote updates. Can I just merge these all then?

adammhaile:
Actually... only the remote json file one really matters for me. Carry on.

PR #305, opened 2017-07-22
Add remote-demo.json

rec:
We should check this in now, and add a unit test for this almost immediately.


adammhaile:
You do realize I'm rewritting remote today and this entire thing will be incorrect, right?

Issue #46, opened 2016-03-25
Unnecessary list surgery in serial_driver.py


rec:
`update` is called on each frame, but you're creating a whole new buffer each time:

https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/drivers/serial_driver.py#L317-L322

That's gotta hurt!  :-)

PR #308, opened 2017-07-23
Remote control is now an animation

adammhaile:
Note: You need the `fixup` branch of BPA to use this because of the pre_run changes. That branch has been updated with new project files too.

A little worse than I thought, but now we could have a remote button that starts a new remote which can start a new remote. So there's that.

I need to make some further passes over this before I'll call it good, but wanted the usual overnight quibbles :P


bibliopixel/animation/animation.py
```diff
@@ -115,6 +115,7 @@ def stamp():
     @contextlib.contextmanager
     def run_context(self):
         self.state = STATE.running
+        self.threading.stop_event.clear()
```
adammhaile: This fixes #306

bibliopixel/animation/collection.py
```diff
@@ -1,11 +1,13 @@
 from .. project import aliases, project
 from . import animation, runner
+import traceback
+from .. import log
 
 
 class Collection(animation.BaseAnimation):
-    def __init__(self, layout, animations=None):
+    def __init__(self, layout, animations=None, no_fail=False):
```
adammhaile: I realized that Collection broke my graceful fail logic that would allow remote to run even if an animation didn't load correctly. So I needed a way to tell _make_animation to not freak out on a failed load.

bibliopixel/animation/collection.py
```diff
@@ -25,9 +27,12 @@ def pre_run(self):
 
     @property
     def current_animation(self):
-        return self.animations[self.index]
+        if self.index >= 0:
+            return self.animations[self.index]
+        else:
+            return None
```
adammhaile: Just seemed like an oversight to me. It would've blown up if you called this before starting.

bibliopixel/animation/collection.py
```diff
@@ -40,4 +45,8 @@ def _make_animation(self, a):
         else:
             desc = {'animation': a[0], 'run': a[1]}
         desc = aliases.resolve(desc)
-        return project.make_animation(layout=self.layout, **desc)
+        try:
+            return project.make_animation(layout=self.layout, **desc)
```
adammhaile: As mentioned above.

bibliopixel/animation/animation.py
rec: Ah, well done.  I had no idea what that was.  How did you figure it out??

bibliopixel/animation/collection.py
rec: This is a great feature, and to be honest, perhaps it should be the default?

bibliopixel/animation/collection.py
rec: This doesn't work for indices greater than the number of animations!

Why not return `self.animations.get(self.index)`?

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+    def __init__(self, layout, animations,
+                 external_access=False, port=5000,
+                 title='BiblioPixel Remote', bgcolor='black',
+                 font_color='white', default=None):
+
+        self.internal_delay = 0  # never wait
+
+        # normalize name and display first
+        for anim in animations:
+            anim['animation']['data'] = dict(DEFAULT_ANIM_CONFIG, **anim['animation'].get('data', {}))
```
rec: Strongly consider extracting `anim['animation']` as e.g. `adesc` as you use it so often.

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            if not anim['animation'].get('name'):
+                raise ValueError('All animations require the `name` parameter: {}'.format(anim['animation']))
```
rec: I hate failing.  Another possibility is to make line 37 `for i, anim in enumerate(animations):' and then change this to:

            if not adesc.get('name'):
                log.error('All animations should have the `name` parameter: %s, adesc)
                adesc['name'] = str(i)

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            if not anim['animation']['data'].get('display'):
```
rec: or `adesc['data']['display'] = adesc['data'].get('display', adesc['name'])`

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+                anim['animation']['data']['display'] = anim['animation']['name']
+            anim['animation']['name'] = ''.join(e for e in anim['animation']['name'] if e.isalnum())
+            anim['run'] = anim.get('run', {})
+            anim['run']['threaded'] = True  # threaded required
+        super().__init__(layout, animations, True)
+
+        self.name_map = {}
+        self.anim_cfgs = []
+        for i in range(len(self.animations)):
```
rec: Lines 50-51 are an antipattern:  replace with `for i, anim in enumerate(self.animations):`

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            anim = self.animations[i]
+            name = animations[i]['animation']['name']  # if failed to load anim will be None
```
rec: OK, now I'm confused - what's the difference between `animations[i]` and `self.animations[i]`??

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            animations[i]['animation']['data']['valid'] = (anim is not None)
+            if name in self.name_map:
+                raise ValueError('Cannot have multiple animations with the same name: ' + name)
+            self.name_map[name] = None if anim is None else i
```
rec: self.name_map[name] = anim and i

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            self.anim_cfgs.append(animations[i]['animation']['data'])
+            self.anim_cfgs[len(self.anim_cfgs) - 1]['name'] = name
```
rec: self.anim_cfgs[-1]['name'] = name

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+            anim.on_completion = self.on_completion
+
+        if default is None:
+            self.default = DEFAULT_OFF
+        else:
+            self.default = ''.join(e for e in default if e.isalnum())
```
rec: This code is repeated above.  Mildly suggest extracting it.  Also suggest using `'_'.join`...

bibliopixel/animation/animation.py
adammhaile: Just by putting some debug statements in there. I quickly realized that what was happening was that after the first go through each animation would run a single frame and then go to the next. I have no idea when exactly this broke. I'd say after 3.0.5 though. I think it worked at some point when I had remote going. I don't know... fixed now :P

bibliopixel/animation/collection.py
adammhaile: :) I debated about that... but if you were running this from a script directly and one of your animations in a sequence failed, then if you missed the traceback printed out to the console you'd probably wonder what happened to that one animation. That was my only concern, but I could go either way on it.

bibliopixel/animation/collection.py
adammhaile: Good catch... yeah, that's better. I always forget that list has a get method.

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
 
-class RemoteControl:
-    def __init__(self, config, animations):
-        # These start with the defaults
-        self.config = dict(DEFAULT_SERVER_CONFIG, **config)
+class RemoteControl(collection.Collection):
+    def __init__(self, layout, animations,
+                 external_access=False, port=5000,
+                 title='BiblioPixel Remote', bgcolor='black',
+                 font_color='white', default=None):
```
adammhaile: Line #46. I call it late because I want to get mangling of names and failing on bad names out of the way first. THEN I call super().

bibliopixel/remote/control.py
adammhaile: oh man... I did not know about that enumerate trick! I always want to get the object and it's position. Very cool.
And yeah... agreed. My other though was to do an inspection and name the animation whatever the class name was...

bibliopixel/remote/control.py
adammhaile: you just hate if statements, don't you?

bibliopixel/remote/control.py
```diff
@@ -14,89 +13,95 @@
...
+        self.internal_delay = 0  # never wait
+
+        # normalize name and display first
+        for anim in animations:
+            anim['animation']['data'] = dict(DEFAULT_ANIM_CONFIG, **anim['animation'].get('data', {}))
+            if not anim['animation'].get('name'):
+                raise ValueError('All animations require the `name` parameter: {}'.format(anim['animation']))
+            if not anim['animation']['data'].get('display'):
+                anim['animation']['data']['display'] = anim['animation']['name']
+            anim['animation']['name'] = ''.join(e for e in anim['animation']['name'] if e.isalnum())
```
adammhaile: No... that would make `Hello!` into `H_e_l_l_o_`
I think what you mean is turn spaces into underscores? I can do that :P

bibliopixel/remote/control.py
adammhaile: As noted above... that's a new trick to me.

bibliopixel/remote/control.py
adammhaile: at this point `animations[i]` is the dict config and `self.animations[i]` is the actual class object.
The reason I did this is that if the animation fails to load, then `self.animations[i]` is None which will have no `name` property. I though about instead of returning None returning like a BaseAnimation instance but that just seemed hackish. But I admit, this is not clear. The BaseAnimation thing would work but I would need to somehow check that that's all it is, therefore it didn't actually load. Thoughts?

bibliopixel/remote/control.py
adammhaile: +1

bibliopixel/remote/control.py
adammhaile: See, I know that indices can wrap around like that, but it annoys me that they do... so I always forget about that trick.

bibliopixel/remote/control.py
adammhaile: Yeah... will do.

bibliopixel/remote/control.py
rec: OOPS!

bibliopixel/remote/control.py
rec: AH, poop, right!

bibliopixel/remote/control.py
rec: :-D  Well, branch prediction, but also people make a lot of conceptual mistakes when reading if statements.

bibliopixel/remote/control.py
rec: `None` is perfect, IMHO.  Which animation did you load?  "None!"

Having a special case for nullable animations is perfectly reasonable, because it's part of your "business logic".

rec:
I agree with everything.  Where do I sign up?

bibliopixel/animation/collection.py
adammhaile: Oh... that's because list does NOT have a get method :P

bibliopixel/remote/control.py
adammhaile: Ahhh... nope, that's no good. Because .get() only chooses your default option if that key DOESN'T EXIST. But in this it does because I always defaulted it to None. In the interest of leaving out the if statement I'm going to instead remove my default to None.

adammhaile:
All fixed. Should be good to go now.

PR #321, opened 2017-07-27
Read long_description from README.md (fix #195)

### LINKED ISSUE #195: Read long_description from README in setup.py
Nice little trick worth noting.  Feel free to close this as "will not fix", it's more for the entertainment value.  :-)

https://github.com/llazzaro/packyou/blob/master/setup.py#L6-L10
https://github.com/llazzaro/packyou/blob/master/setup.py#L28

rec:


rec:
Cool, now I understand this better.  I repro'ed the issue with the original commit, then made the fix, tested, and it works now.

Question: why doesn't Travis catch this?

adammhaile:
Because Travis doesn't try to install from an sdist... which we should
probably change.

rec:
So do you really need the transcript from me or are you willing to believe that I actually tested it?  :-D

adammhaile:
LOL... sorry. Crazy week. We're good here. +1 Merge away.

PR #325, opened 2017-07-27
Add `.` to `sys.path` in `bibliopixel run` (fix #323)

### LINKED ISSUE #323: pwd should always be on the path
This is kind of in the interest of eventually only really supporting running BP from `bibliopixel run`, no direct scripting.... whaaaaaaaaat? Yes, you heard me.
If I call bibliopixel run, whatever directory I'm currently in should be on the python import path. Right now when I dev I have to manually add that path and it's annoying.
Better yet, whatever directory the json file is in (with a few blacklisted exceptions or files loaded from URL) should be on the path too. So if I call `bibliopixel run ./MyAwesomeDemo/config.json` and `MyAwesomeDemo` has my code too, it should just plain work.

rec:


bibliopixel/main/common_flags.py
```diff
@@ -105,3 +105,7 @@ def get_value(name):
         project_flags['maker'] = {'use_numpy': True}
 
     return project.project_to_animation(desc, project_flags)
+
+
+def extend_path(args):
+    gitty.sys_path.extend((args.path + ':.') if args.path else '.')
```
adammhaile: Resolve the fully qualified path with os.getcwd()
Unless gitty is doing some magic here then all that does is add wherever you already are to the python path, but means that if you change dirs, then THAT is in your path, not the old folder. That's kinda the point of this... importing via the project file *should* work for files already in your PWD, but it doesn't always and adding the fully qualified path to your python path fixes that.

rec:
Done and tested.

PR #328, opened 2017-07-29
Project refactoring

rec:
These changes are on the way to the correct "default driver/layout/etc" idea.

There are two things going on here!

First is that my whole idea of "settings" is wrong.  So I killed it.  :-D  We'll be using just `DataFile` and life will be much simpler.

Then before I went ahead, I cleaned up project.py [considerably](https://github.com/ManiacalLabs/BiblioPixel/compare/dev...rec:project-refactoring#diff-398a55c2da42aa3c07fa0ba2a0f1edde) and fixed a few other minor issues.


bibliopixel/animation/animation.py
```diff
@@ -138,15 +138,15 @@ def run_all_frames(self):
                 self.run_one_frame()
 
     def set_runner(self, runner):
-        self.runner = runner
+        self.runner = Runner(**(runner or {}))
```
adammhaile: Why? Weren't we passing in a complete Runner() before?

bibliopixel/animation/animation.py
```diff
@@ -138,15 +138,15 @@ def run_all_frames(self):
...
         self.threading = AnimationThreading(self.runner, self.run_all_frames)
 
     def start(self):
         self.threading.start()
 
     def run(self, **kwds):
         # DEPRECATED
-        self.set_runner(Runner(**kwds))
+        self.set_runner(kwds)
```
adammhaile: Oh... I see. Still, why the change?

bibliopixel/layout/matrix.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, drivers, width=0, height=0, coordMap=None,
         driver - instance that inherits from DriverBase
         width - X axis size of matrix
         height - Y axis size of matrix
-        coordMap - a 2D matrix defining the X,Y to strip index mapping. Not needed in most cases
-        rotation - how to rotate when generating the map. Not used if coordMap specified
+        coord_map - a 2D matrix defining the X,Y to strip index mapping. Not needed in most cases
+        rotation - how to rotate when generating the map. Not used if coord_map specified
```
adammhaile: You're breaking 3.0.5 compat... I mean, yeah, no one probably uses it directly, just wanted to note it.

bibliopixel/layout/multimap.py
```diff
@@ -1,10 +1,33 @@
...
+    def _make_driver(self, width, height, matrix=None, **kwds):
+        row = gen_matrix(width, height, **(matrix or {}))
+        for y, row_entry in enumerate(row):
+            self.map.append([i + self.offset for i in row_entry])
+
+        self.offset += len(row) * len(row[0])
+        return self.make_object(width=width, height=height, **kwds)
 
     def addRow(self, *maps):
+        # DEPRECATED
```
adammhaile: Why?

bibliopixel/project/project.py
```diff
@@ -1,67 +1,54 @@
...
+        raise ValueError('layout was not specified in project')
+
+    if not (driver or drivers):
+        raise ValueError('Projects has neither driver nor drivers sections')
+
+    gitty.sys_path.extend(path or '')
+    maker = data_maker.Maker(**(maker or {}))
+    make_object = functools.partial(importer.make_object, maker=maker)
+
+    builder = MultiMapBuilder(make_object)
```
adammhaile: We still need to make all this easier in project files.

bibliopixel/animation/animation.py
rec: Why should other classes need to know about the existence of Runner, which they never see?

bibliopixel/layout/matrix.py
rec: Yes, that's true.  :-/

bibliopixel/layout/multimap.py
rec: `add_one_row` completely replaces `addRow`.

In practice `len(maps)` is always 1. Nearly all that code is involved with keeping track of the cases where `len(maps) > 1`.

(More, calling `add_one_row` repeatedly has exactly the same effect that `addRow` would have had with `maps` of length greater than 1.  !)

bibliopixel/project/project.py
rec: Agree 100% - it's also necessary for #270.

bibliopixel/animation/animation.py
adammhaile: Fair enough.

bibliopixel/layout/multimap.py
adammhaile: +1

Issue #329, opened 2017-07-29
Now we have `.` in the path, scripts aren't necessarily exportable...


rec:
...because they can refer to local functions that aren't visible elsewhere!

adammhaile:
Not everyone is going to have all their code up in a git repo somewhere. This is a non-issue as it's still valid even if not ideal for sharing with the community.

rec:
Absolutely it's an issue, because we need a way to prevent people from doing this when they _do_ upload!

Basically, as it works now, people will prepare their files and they will fail the first time for everyone.  Bad show.

It's not just `.`, it's also local files for datafiles.

Still thinking about this one... I think we'll eventually need a "validate local" or something.

adammhaile:
As noted previously we need a distribute command that ensures a file is
globally usable.

rec:
On it.  :-D

PR #330, opened 2017-07-29
New --external flag.

rec:


rec:
I don't like `--external` but `--sanity-check` doesn't seem really super.

What are we doing...?  We're checking to see - if we can share this code.

`--check_shareable`

Urg, I just realized that I need to add something else... moment pls!

rec:
I am now also checking the path passed to gitty.

Got a better name?

    --check_shareable
    --check_push

???

adammhaile:
Check dist?

rec:
My final thought on this is that we won't come up with a final name until we have names for the whole upload/sharing procedure, so I'm going to squash and push, and then we'll change the name when we've unified our nomenclature.

rec:
Fixes #329.

PR #339, opened 2017-08-01
Remote UI Triggers

adammhaile:
I fully expect a few semantic quibbles here, but the code works and is sound.
First note that you need to clone and install this: https://github.com/ManiacalLabs/BiblioPixelTriggers
That includes what is currently the only available trigger, `crontab`, and I intend to update BP to install BPT by default. BPT, however, will NOT install any dependencies except for `croniter` used by `crontab` as it's simple and I want `crontab` to be always available. When I add more triggers like Bluetooth proximity, I will of course just have it gracefully fail and tell you how to install the dependencies. This is why I actually import the module once first in control.py and then later in trigger_process.py - I want it to fail before it would fail inside a new process.

On that note... The remote animation now takes a `triggers` param like this:
```
 "triggers": [
            {
                "typename": "BiblioPixelTriggers.crontab.crontab",
                "config": "*/2 * * * *",
                "name": "Bloom"
            },
            {
                "typename": "BiblioPixelTriggers.crontab.crontab",
                "config": "1-59/2 * * * *",
                "name": "OFF_ANIM"
            }
        ]
```
The Remote animation, however, will create only a single process for each *type* of trigger, `crontab` in this instance. The reason for this layout is to have one entry per trigger, even if there are multiple triggers that are all the same type. This decreases the total possible depth of the `triggers` dict. But what I do is bundle all config/name pair for each trigger type and pass them as a list to that trigger class. So that trigger type gets one process and one class instance and it's expected to handle that list of config/name pairs appropriately.

One thing I need to resolve is aliases for the trigger typename... though maybe not. Since triggers live in a different repo and PyPi package we'd have to keep things in sync, which I'd rather not.

I probably should also do that little hack that makes a module become a sub-object... basically `BiblioPixel.crontab` instead of `BiblioPixel.crontab.crontab`... need to look up how to do that again. But that would at least decrease typename length.

Demo project file (GitHub would not let me attach a json file for some reason).
This will run Bloom on even minutes and OFF_ANIM (which is MatrixRain in this case) on odd minutes. `crontab` supports everything that actual `crontab` does. I was going to make my own kind of close version, but then found `croniter` and got lazy, deciding that this was perfect for what I needed.
```
{
    "animation": {
        "title": "Trigger Demo",
        "typename": "remote",
        "animations": [
            {
                "animation": {
                    "typename": "BiblioPixelAnimations.matrix.MatrixRain.MatrixRain",
                    "name": "MatrixRain"
                },
                "run": {
                    "fps": 30
                }
            },
            {
                "animation": {
                    "typename": "BiblioPixelAnimations.matrix.bloom.Bloom",
                    "name": "Bloom"
                },
                "run": {
                    "fps": 30,
                    "amt": 6
                }
            },
            {
                "animation": {
                    "typename": "BiblioPixelAnimations.matrix.perlin_simplex.PerlinSimplex",
                    "name": "PerlinSimplex"
                },
                "run": {
                    "fps": 30
                }
            }
        ],
        "default": "MatrixRain",
        "bgcolor": "black",
        "external_access": true,
        "port": 5000,
        "triggers": [
            {
                "typename": "BiblioPixelTriggers.crontab.crontab",
                "config": "*/2 * * * *",
                "name": "Bloom"
            },
            {
                "typename": "BiblioPixelTriggers.crontab.crontab",
                "config": "1-59/2 * * * *",
                "name": "OFF_ANIM"
            }
        ]
    },
    "driver": {
        "num": 1024,
        "typename": "simpixel"
    },
    "layout": {
        "typename": "matrix",
        "width": 32,
        "height": 32
    }
}
```

### DIFF
--- a/bibliopixel/remote/control.py
+++ b/bibliopixel/remote/control.py
@@ -2,7 +2,10 @@ import multiprocessing
 from .. animation.animation import STATE
 from .. animation import collection
 from . import server
+from . import trigger_process
 from .. util import log
+from .. util import importer
+import copy
 
 
 DEFAULT_OFF = 'OFF_ANIM'
@@ -32,7 +35,8 @@ class RemoteControl(collection.Collection):
     def __init__(self, layout, animations,
                  external_access=False, port=5000,
                  title='BiblioPixel Remote', bgcolor='black',
-                 font_color='white', default=None):
+                 font_color='white', default=None,
+                 triggers=[]):
 
         self.internal_delay = 0  # never wait
 
@@ -47,10 +51,12 @@ class RemoteControl(collection.Collection):
             adesc['name'] = normalize_name(adesc['name'])
             anim['run'] = anim.get('run', {})
             anim['run']['threaded'] = True  # threaded required
-        super().__init__(layout, animations, True)
+
+        super().__init__(layout, copy.deepcopy(animations), True)
 
         self.name_map = {}
         self.anim_cfgs = []
+
         for i, anim in enumerate(self.animations):
             adesc = animations[i]['animation']
             name = adesc['name']  # if failed to load anim will be None
@@ -88,13 +94,16 @@ class RemoteControl(collection.Collection):
             'title': title
         }
 
-        self.q_send = multiprocessing.Queue()
         self.q_recv = multiprocessing.Queue()
+
+        self.send_queues = {
+            'RemoteServer': multiprocessing.Queue()
+        }
 
         server_args = (
             external_access,
             port,
-            self.q_send,
+            self.send_queues['RemoteServer'],
             self.q_recv
         )
 
@@ -104,13 +113,32 @@ class RemoteControl(collection.Collection):
         self.handlers = {
             'run_animation': self.run_animation,
             'stop_animation': self.stop_animation,
-            'get_config': self.get_config
+            'get_config': self.get_config,
+            'trigger_animation': self.run_animation
         }
+
+        self.trigger_procs = {}
+        self.triggers = {}
+        for trigger in triggers:
+            typename = trigger.pop('typename')
+            if typename:
+                importer.import_symbol(typename)  # attempt early to fail early
+                self.triggers.setdefault(typename, []).append(trigger)
+            else:
+                raise ValueError('Triggers require a `typename` field!')
+
+        for typename, configs in self.triggers.items():
+            self.trigger_procs[typename] = multiprocessing.Process(
+                target=trigger_process.run_trigger,
+                args=(typename, self.q_recv, configs))
 
     def cleanup(self, clean_layout=True):
         self.q_recv.close()
-        self.q_send.close()
+        for q in self.send_queues.values():
+            q.close()
         self.server.terminate()
+        for _, t in self.trigger_procs.items():
+            t.terminate()
         super().cleanup(clean_layout)
 
     def on_completion(self, reason):
@@ -146,6 +174,8 @@ class RemoteControl(collection.Collection):
 
     def pre_run(self):
         self.server.start()
+        for proc in self.trigger_procs.values():
+            proc.start()
 
     def step(self, amt=None):
         self.run_animation()
@@ -156,4 +186,7 @@ class RemoteControl(collection.Collection):
                 resp = False, '{} is not a valid request!'.format(req)
             else:
                 resp = self.handlers[req](recv['data'])
-            self.q_send.put(resp)
+
+            resp_q = self.send_queues.get(recv['sender'])
+            if resp_q:
+                resp_q.put(resp)
--- a/bibliopixel/remote/trigger_process.py
+++ b/bibliopixel/remote/trigger_process.py
@@ -1,0 +1,10 @@
+from .. util import importer
+
+
+def run_trigger(typename, q, configs):
+    trigger_class = importer.import_symbol(typename)
+    trigger = trigger_class(q, configs)
+    try:
+        trigger.start()
+    except KeyboardInterrupt:
+        pass
--- a/bibliopixel/project/project.py
+++ b/bibliopixel/project/project.py
@@ -15,14 +15,15 @@ Your path was %s."""
 def make_animation(layout, animation, run=None):
     animation = aliases.resolve(animation)
     reserved = {p: animation.pop(p, None) for p in RESERVED_PROPERTIES}
-    animation = importer.make_object(layout, **animation)
+    animation_obj = importer.make_object(layout, **animation)
 
     # Add the reserved properties back in.
     for k, v in reserved.items():
-        (v is not None) and setattr(animation, k, v)
+        animation[k] = v
+        (v is not None) and setattr(animation_obj, k, v)
 
-    animation.set_runner(run)
-    return animation
+    animation_obj.set_runner(run)
+    return animation_obj
 
 
 def extend_path(path):
--- a/bibliopixel/remote/server.py
+++ b/bibliopixel/remote/server.py
@@ -1,5 +1,6 @@
 import logging, os, queue, sys
 from .. util import log
+import flask
 
 MESSAGE = """\
 Remote UI Server available at: {1}
@@ -17,7 +18,6 @@ class RemoteServer:
         cdir = os.path.dirname(os.path.realpath(__file__))
         static_dir = os.path.abspath(os.path.join(cdir, '../../ui/web_remote'))
 
-        import flask
         self.app = flask.Flask('BP Remote', static_folder=static_dir)
 
         self._set_routes()
@@ -45,14 +45,13 @@ class RemoteServer:
         return self.api('stop_animation')
 
     def api(self, request, data=None):
-        self.q_send.put({'req': request.lower(), 'data': data})
+        self.q_send.put({'req': request.lower(), 'data': data, 'sender': 'RemoteServer'})
 
         try:
             status, data = self.q_recv.get(timeout=5)
         except queue.Empty:
             status, data = False, 'Timeout waiting for response.'
 
-        import flask
         return flask.jsonify({
             'status': status,
             'msg': 'OK' if status else data,
--- a/bibliopixel/remote/trigger.py
+++ b/bibliopixel/remote/trigger.py
@@ -1,0 +1,10 @@
+from .. util import importer
+
+
+class TriggerBase:
+    def __init__(self, q, configs):
+        self.q = q
+        self.configs = configs
+
+    def trigger(self, name):
+        self.q.put({'req': 'trigger_animation', 'data': name, 'sender': 'Trigger'})


rec:
Super cool, but must wait till tomorrow for the review... 😃

rec:
So I'm reading this again now, and again, it's super-cool!

I might not get a chance to actually read the code till tonight, but here are some thoughts.

---

First, aliases are a continuing pain-point, but super-useful.  

Two basic, intertwined issues: once someone gets going, most of the animations they use will not be from BP, so aliases will be useless.

So you naturally want the ability to use your own aliases - but then this results in a project that isn't universally usable.

Also, #340 also comes into play.  An alias on its own cannot represent fully an external class, without an additional entry in the `path` component of a project, which makes aliases to external classes mostly useless.

I think the --external command will come to the rescue here, but the details are still developing.

---

Second, the idea of a trigger being a typename is particularly strong (of course, I'm bound to like it :-D).  The only potential beef is the definition of `config` which is basically a constructor - but one that right now can only take a string.

In particular, crontab entry format has proven extremely problematic to people in the past, because it's hard to specify and hard to test without sitting around waiting.

More than twice in my life, things went wrong because some engineer who was smart enough to know better, sometimes not me, made a mistake in configuring a crontab.

I don't think you should change that for this commit - but I think we should be moving towards a human-readable way to construct crontab entries, _and_ some way to test them without waiting (that is pretty easy - just two commands, one that says, "Print out the times that this crontab entry will be executed" and the other that says, "Execute all the commands between now and this time T in the future").

In particular, the crontab mechanism needs to be expansible, so we can do things like, "Turn this on at dusk".

Reliable, easy time specifications are what's going to make this sing.  Now we're getting into this, we need to be extremely thoughtful of the needs of our users.

Now off to finish fixing my wife's bike.

adammhaile:
On Wed, Aug 2, 2017 at 7:03 AM, Tom Swirly <notifications@github.com> wrote:
>
> Also, #340 <https://github.com/ManiacalLabs/BiblioPixel/issues/340> also
> comes into play. An alias on its own cannot represent fully an external
> class, without an additional entry in the path component of a project,
> which makes aliases to external classes mostly useless.
>
> I think the --external command will come to the rescue here, but the
> details are still developing.
>
Agreed. And I have no desire or intention to solve this for this release. I
just wanted to note that I had thought about it.

> ------------------------------
>
> Second, the idea of a trigger being a typename is particularly strong (of
> course, I'm bound to like it :-D). The only potential beef is the
> definition of config which is basically a constructor - but one that
> right now can only take a string.
>
No no, config can be whatever your trigger class wants. It's just that for
`crontab` it only needs a string.

> I don't think you should change that for this commit - but I think we
> should be moving towards a human-readable way to construct crontab entries,
> *and* some way to test them without waiting (that is pretty easy - just
> two commands, one that says, "Print out the times that this crontab entry
> will be executed" and the other that says, "Execute all the commands
> between now and this time T in the future").
>
Yeah, that's what I started with. And it exploded into this huge
complicated thing... then I remembered `croniter` which makes my trigger
code super simple as I can just call `get_next()` and not worry about the
rest. I plan on also adding an installed script to the installation of BPT
that you can basically do `crontab-test -n 20 "*/5 * * * *"` and it will
print the next 20 times that match your crontab. I will also link to here
in the docs: https://crontab.guru/

But yes... this is V1, the version that satisfies my need right now with
the least possible work. I think that in the future we can add a more
simplistic `BiblioPixelTriggers.DateTime` or something which was my
original plan that allows something more human readable. Really what I was
doing with `crontab` was making something that did just enough that I could
figure out the BP side of things. Now we have something that is generic and
should work with any type of trigger we want to throw at it.

> In particular, the crontab mechanism needs to be expansible, so we can do
> things like, "Turn this on at dusk".
>
Hold yer horses buddy, that one is coming and will either be it's own
trigger or will be part of the future more friendly time/date/sun trigger.
I'm already working on a trigger that calculates sun rise/set and all the
variances of that (civil & nautical twilight and such) and then lets you
base things off that. I'm going to use it for my porch lights.
Again, `crontab` is the first proof of concept. My plan is to release this
in BP 3.2 or whatever and then build out BPT separately (which is why I
broke it out).... and THEN announce it down the road when there are more
triggers.

> Reliable, easy time specifications are what's going to make this sing. Now
> we're getting into this, we need to be extremely thoughtful of the needs of
> our users.
>

Agreed.... but note above. This is just enough to get the BP side of things
working. More and better triggers can and will come later. So don't worry
about that.

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
         self.handlers = {
             'run_animation': self.run_animation,
             'stop_animation': self.stop_animation,
-            'get_config': self.get_config
+            'get_config': self.get_config,
```
rec: Since these are growing, perhaps we should automate this a bit:

    HANDLER_NAMES = 'get_config', 'run_animation', 'stop_animation', 'trigger_animation', 
    
    def __init__(self, ...):
        self.handlers = {k: v for k, v in self.__dict__ if k in self.HANDLER_NAMES}

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
...
+            typename = t['typename']
+            importer.import_symbol(typename)  # attempt early to fail early
+            if typename not in self.triggers:
+                self.triggers[typename] = []
+            t.pop('typename')
+            self.triggers[typename].append(t)
+
+        for typename, configs in self.triggers.items():
+            trigger_args = (typename, self.q_recv, configs)
+            self.trigger_procs[typename] = multiprocessing.Process(target=trigger_process.run_trigger,
```
rec: Does this have to be a process?

Since you're using your own queue, I could simplify `bibliopixel.threads.sub.run` to your needs here, perhaps in a later pull request - then you could switch between threading and mp.

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
...
+                                                                   args=trigger_args)
```
rec: This line is super-long, what about:

    self.trigger_procs[typename] = multiprocessing.Process(
        target=trigger_process.run_trigger,
        args=(typename, self.q_recv, configs))

and dump the uninformative local variable.

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
...
             'stop_animation': self.stop_animation,
-            'get_config': self.get_config
+            'get_config': self.get_config,
+            'trigger_animation': self.run_animation
         }
 
+        self.trigger_procs = {}
+        self.triggers = {}
+        for t in triggers:
+            typename = t['typename']
```
rec: What happens if `typename` isn't in the triggers?  This is an error that can be caused by bad user data, so it shouldn't cause the program to crash!

You should report this and then, probably, just skip that trigger rather than failing.

bibliopixel/remote/trigger_process.py
```diff
@@ -0,0 +1,10 @@
+from .. util import importer
```
rec: I suggest you move all of this file into remote/trigger.py - they are both tiny.

bibliopixel/remote/control.py
```diff
@@ -88,13 +91,16 @@ def __init__(self, layout, animations,
             'title': title
         }
 
-        self.q_send = multiprocessing.Queue()
         self.q_recv = multiprocessing.Queue()
 
+        self.resp_qs = {
```
adammhaile: ha... yeah, send_queues is fine. I don't know what I was thinking there.

bibliopixel/remote/control.py
adammhaile: Hahaha... I originally had that. And the reason I didn't do that here is that trigger_animation just calls run_animation... the reason I did this is run_animation pushes back to a send queue, trigger_animation does not. I didn't want the triggers have to care about getting a response from a queue, so that's why there's all the mangling of do I send a response or not?

bibliopixel/remote/control.py
adammhaile: Yeah... I debated about that. I'm kinda keen on this hard failing because they may not see the message and then wonder why their trigger isn't happening. But I can certainly improve the hard fail error message.

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
...
+        self.trigger_procs = {}
+        self.triggers = {}
+        for t in triggers:
+            typename = t['typename']
+            importer.import_symbol(typename)  # attempt early to fail early
+            if typename not in self.triggers:
+                self.triggers[typename] = []
+            t.pop('typename')
+            self.triggers[typename].append(t)
+
```
adammhaile: Fair enough.

bibliopixel/remote/control.py
adammhaile: I don't want the triggers to in any way mess with the performance of the animation. `crontab` wouldn't do that now but I have plans for some that might. And getting them out of the GIL and onto another process was the only good way to do that.
I would prefer that I keep my current method now and if you implement core mp functionality later I can switch over to that.

bibliopixel/remote/control.py
adammhaile: I can never get the indentation write when I do that (new line after the `(` )... But yeah, I'm fine with that.

bibliopixel/remote/control.py
```diff
@@ -104,13 +110,32 @@ def __init__(self, layout, animations,
...
+            self.trigger_procs[typename] = multiprocessing.Process(target=trigger_process.run_trigger,
+                                                                   args=trigger_args)
+
     def cleanup(self, clean_layout=True):
         self.q_recv.close()
-        self.q_send.close()
+        for q in self.resp_qs.values():
+            q.close()
         self.server.terminate()
+        # for _, t in self.trigger_procs.items():
```
adammhaile: Oh, shit... that was supposed to be actually used. I commented it for testing something. It needs to go back in.

bibliopixel/remote/control.py
```diff
@@ -146,6 +171,8 @@ def get_config(self, data):
 
     def pre_run(self):
         self.server.start()
+        for typename, proc in self.trigger_procs.items():
```
adammhaile: Yeah... I had a need for typename before then removed it. Will change.

bibliopixel/remote/trigger_process.py
adammhaile: Can't... the process method has to be in it's own module to work on Windows because... Windows. And yes... I'm keeping the process for now :)

rec: Hah, madness!  :-D 

The process is fine.  I just wanted to make sure you were sure.

rec:
👍 if you make the changes you promised.  :-D

bibliopixel/remote/trigger_process.py
adammhaile: :) Yeah... Windows kicks my ass here. I'd love to drop it from support but that would kinda defeat the point of Python I guess :P

adammhaile:
Cool... will do tonight.

PR #342, opened 2017-08-03
Better aliases (fix #339)

rec:
Voilà, we have:

* got rid of the three sections for aliases.
* allowed the `$` prefix for alias expansion of segments of typenames.
* and there are now user-definable aliases!
* which can be set, listed and delete from `bibliopixel`.
* and which respect `--external` (renamed to `--isolate`)!


rec:
Take your time.  I'm doing this to delay doing other less-fun things.  :-D

PR #347, opened 2017-08-05
Added brightness slider

adammhaile:


bibliopixel/layout/layout.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, drivers, threadedUpdate, brightness,
         self.animation_sleep_time = None
 
         self.threading = UpdateThreading(threadedUpdate, self)
+        self.brightness = brightness
```
adammhaile: There was nothing in layout storing this... only in Drivers. I honestly don't care what it is in the drivers, but I care about what the layout *thinks* it is.

bibliopixel/remote/control.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, layout, animations,
             name = adesc['name']  # if failed to load anim will be None
             if anim is None:
                 adesc['data']['valid'] = False
-                adesc['data']['display'] = 'LOAD FAILED: ' + adesc['data']['display']
```
adammhaile: Took up too much space.

adammhaile:
Indented to be super simple, graphically. I'm not a designer so I could not decide on a good way to actually label this... I'm open to suggestions. But my CSS-fu sucks, so there's not a lot I can do.
Just tap or drag on the bar on the left to set the brightness level.
Note, the API now also has this option http://localhost:5000/api/brightness/64

bibliopixel/layout/layout.py
rec: This is reasonable, but only because one layout can have more than one driver.

bibliopixel/remote/control.py
```diff
@@ -114,7 +114,8 @@ def __init__(self, layout, animations,
             'run_animation': self.run_animation,
             'stop_animation': self.stop_animation,
             'get_config': self.get_config,
-            'trigger_animation': self.run_animation
```
rec: This line is in the delta because you won't leave trailing commas on your lists and dicts - which is best practice for this reason!

bibliopixel/remote/control.py
```diff
@@ -165,9 +166,20 @@ def run_animation(self, name=None):
     def stop_animation(self, data):
         return self.run_animation()
 
+    def change_brightness(self, data):
+        if isinstance(data, str):
+            try:
+                data = int(data)
+            except:
+                return False, 'Invalid brightness value!'
```
rec: Why not just throw an exception here and have the parent class catch it?  That first `True` vs `False` is what you'd do in a language without exceptions!

bibliopixel/remote/control.py
adammhaile: ah...good to know. No other language I use allows them, so it's just habit.

bibliopixel/remote/control.py
adammhaile: Exceptions are expensive. They are for when you can proceed no longer. The thing is that here I don't really care that I got a bad value, I'm just informing the idiot that passed in `twelve` for a value.
Yes, there are a few rare times when I'll throw an exception, handle it, and keep going... but exception != return error. Especially in an REST interface.

adammhaile:
The .min.js files *did* completely change in this commit. I used to use zepto and switch to jquery. Every js file that wasn't main.js is completely new in this commit.

adammhaile:
Got another one incoming, so merging... if there's any thing egregious we can always fix it later.

PR #353, opened 2017-08-08
Simpler API for reading projects.

rec:
Note that I don't actually call `start` - that's pretty dramatic, you might want to actually do more stuff before the project starts.

gitty.raw.request now gets local files as well as URLs.


adammhaile:
Nice that was fast! :) +1

PR #368, opened 2017-08-15
Convert README to be an rst document (fix #195)

### LINKED ISSUE #195: Read long_description from README in setup.py

rec:


rec:
https://pandoc.org/try/ was the key URL.

PR #372, opened 2017-08-15
Fix three threading issues.

rec:


bibliopixel/animation/runner.py
```diff
@@ -39,4 +40,4 @@ def __init__(self, amt=1, fps=None, sleep_time=0, max_steps=0,
         self.max_cycles = max_cycles
         self.seconds = seconds
         self.threaded = threaded
-        self.join_thread = join_thread
```
adammhaile: So... how do you join the thread now?

bibliopixel/threads/animation_threading.py
```diff
@@ -50,13 +50,23 @@ def wait(self, wait_time, timestamps):
...
+            self.thread.start()
 
-        self.thread = threading.Thread(target=self.target, daemon=True)
-        self.thread.start()
+        if self.runner.main:
+            if self.runner.threaded:
+                start_thread(self.target)
+                self.runner.main()
 
-        if self.runner.join_thread:
```
adammhaile: Ah...

PR #384, opened 2017-08-19
Megatest - part i

rec:


adammhaile:
Good stuff... I'm about to push 3.3 soon though and this is going to have to hang out till after that.

rec:
Ah, great, I'll remerge the code into all_test.py then while we're waiting!  :-)

rec:
So this is less ready than when I submitted it.  :-D

It needs a merge, and plus I realized I need better error handling (see #385) - and there's an actual bug that opens as many copies of SimPixel as there are projects :-D - but it's really neat.

You can `bp foo.json bar.json baz.json --run_for=3` and run each of these animations for three seconds (if they aren't a remote) and then stop.

Or if you `bp foo.json bar.json baz.json`, it will run each animation for as long as it takes - or until you hit control C, when it jumps to the next animation(!).  (As long as none of the animations are a remote.  :-D)

The unit tests now smoke test BPA, and I'm going to check in some projects that test each layout successively.

And this means that I will have reduced the "megatest" to two steps: "run unittests" as before, and `bp test/projects/*` - which means, I think, that I'm not even going to bother to check in a script to do those two steps.

!

rec:
This pull request ended up being decomposed into the other 9 on this queue.

PR #391, opened 2017-08-22
New colors type for projects

rec:


bibliopixel/project/types/colors.py
```diff
@@ -0,0 +1,22 @@
...
+
+@functools.singledispatch
+def make(c):
+    raise ValueError("Don't understand type %s" % type(c), USAGE)
+
+
+@make.register(tuple)
+@make.register(list)
+def _(c):
+    return [color.make(i) for i in c]
```
adammhaile: Don't quite understand how color.make() works as make(c) is also just a `raise ValueError` in that module. What is it actually calling?

adammhaile:
Comment was just a question, but I trust it works... merging.

bibliopixel/project/types/colors.py
rec: It's `singledispatch`!

https://docs.python.org/3/library/functools.html#functools.singledispatch

A really nice facility, but rarely the correct solution to a problem.  However, this one problem it's perfect for (and is 90%+ of the time I've seen it used...)

PR #392, opened 2017-08-23
Collection/Sequence could start next before dying

adammhaile:
@rec This was a fun one to figure out. When connected to a display that's a little slow to setup, a sequence had a non-zero chance that when cleanup() was called, the animation would loop back around again and start the next animation before completely dying. It then put it into a state of weirdness where you could not kill it anymore because it thought it was already dead. I put the fix in collection because this will and should affect anything that derives from collection. When cleanup is called, state should always be set so nothing else happens.


adammhaile:
Tiny tiny thing. And trying to get 3.3.1 out. Merging.

PR #394, opened 2017-08-23
Fix indentation in matrix.py

rec:
flake8 complains that this block is badly indented.  Not sure why it isn't stopping Travis...?


adammhaile:
Oh...um, wow. I know why. Because travis doesn't run Flake8!
I was kind of shocked myself. All it runs is the coverage test.
I'm sure I meant to run flake8 but it never has. Ever.
So, may I request you ammend this commit to include `flake8 .` in the travis scripts?
Then I'll pull it in for the 3.3.1 release.

rec:
I turned on flake8.  I also tested that it worked, by simply pushing the flake8 change without the indentation change, and travis let me know about it really fast!

PR #396, opened 2017-08-23
New `bp` update flag `--run_for`

rec:
These are all little reviews spinning off the big "megatest" review.


adammhaile:
Grr... the travis fail is that damn threading test again that I made a bug for. Drives me crazy since if you just rerun that one 3.6 part of that build it will be fine again.
Something is up with that test.

PR #406, opened 2017-08-26
Rotation becomes a number.

rec:
I woke up this morning and went DOH.

There's a much easier solution, and that involves the rotation enum having the same value as its decimal degrees, so `Rotation.ROTATE_180` is `180`!

I made the change pretty fast (and the whole code path is heavily tested, so I'm confident it's correct) but when I saw the result, I think we can do even better:

* Get rid of the whole `Rotation` enum!
* Make `rotation` just an `int`.
* When rotation must be a multiple of 90º and it isn't (which is exactly in one place, the constructor of `layout.matrix.Matrix`), warn the user and round to the nearest multiple of 90º

I didn't go on because I didn't want to invest time without seeing what you thought, but I think it's a much better solution.

Less code;  less to explain to people;  we can warn people and "do a good job" by rounding instead of failing.

FULL DISCLOSURE.

There's one additional line of somewhat obscure code added [here](https://github.com/rec/BiblioPixel/blob/rotation/bibliopixel/layout/geometry/rotation.py#L17).

However, this obscurity was there in the original code - it was the fact that `ROTATE_90, ROTATE_180, ROTATE_270` was `3, 2, 1` instead of `1, 2, 3`.  At least in the new code, it's obvious that something was going on there, whereas in the original, it's hidden.

So if you give me the go-ahead, I can go all the way with this...


rec:
Nah, I need to close this and reopen it again `master` anyway.

I'll redo it with that whole enum gone send it to you!

adammhaile:
No. I'm not taking this on master. That can wait. Too much of a change.

rec:
great, I reopened it anticipating this.  :-D

I'll fix it up and let you know.

adammhaile:
Lol..  yeah. Just merge on dev. Not a problem

rec:
All right, please take a quick look.

I added a test for the new warning functionality in project_test.py, and it was good that I did, because I managed to make three errors in doing it.  :-o

So I'm pretty sure it works.

-55 lines today, and -10 more when we hit v4.0 (deprecations).  Good antiwork!

adammhaile:
+1

PR #407, opened 2017-08-26
Clear numpixel arrays before using them.

rec:
This is the change that makes numpy work with almost perfect backward and forwards compatibility with lists of tuples.

(Again, lists of tuples are still the default, but you can change it with a command line argument...)

I was able to run all of the strip tests but one, and all of the matrix tests but one, and all of the cube and circle tests and remote tests right out of the box!

(from here: https://github.com/ManiacalLabs/BiblioPixel/tree/dev/projects)

I'll send out a review for the two one-line BPA changes needed to support this.  They are of course backward compatible.


rec:
Heh, just saw this!  :-D

Sure, this is zero issue to me - the actual payload change is [tiny](https://github.com/ManiacalLabs/BiblioPixelAnimations/pull/29/commits/d479ae840353b3b187da6c9d16e03dda81fd4015) and can be checked in w/e.

PR #409, opened 2017-08-28
Make sure all aliases exist (fix #408)

### LINKED ISSUE #408: networkudp alias is wrong
https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/project/alias_lists.py#L17
Should be `bibliopixel.drivers.network_udp.NetworkUDP`

rec:


bibliopixel/project/alias_lists.py
```diff
@@ -7,18 +7,17 @@
 
 BUILTIN_ALIASES = {
     # drivers
-    'apa102': 'bibliopixel.drivers.API.APA102.APA102',
```
adammhaile: Whaaaat?! How the hell did we miss that one! Well, other than the fact that I always use `spi` with `ledtype`

bibliopixel/project/alias_lists.py
```diff
@@ -7,18 +7,17 @@
...
+    'apa102': 'bibliopixel.drivers.SPI.APA102.APA102',
+    'sk9822': 'bibliopixel.drivers.SPI.APA102.APA102',
     'dummy': 'bibliopixel.drivers.dummy_driver.Dummy',
     'hue': 'bibliopixel.drivers.hue.Hue',
-    'image': 'bibliopixel.drivers.image_sequence.ImageSequence',
-    'lpd8806': 'bibliopixel.drivers.API.LPD8806.LPD8806',
+    'lpd8806': 'bibliopixel.drivers.SPI.LPD8806.LPD8806',
     'network': 'bibliopixel.drivers.network.Network',
-    'network_udp': 'bibliopixel.drivers.network.NetworkUDP',
+    'network_udp': 'bibliopixel.drivers.network_udp.NetworkUDP',
```
adammhaile: Clearly you didn't rebase in master as I already fixed this in v3.3.3

rec:
Rebased.

We missed all of these because I did not write a test with this code.  

The "lack of test" indicates that I didn't think this through.

Though perhaps we didn't have the import machinery at that time.

I looked at the fact that none of these were tested and I guessed there would be two other mistakes and there were three when I actually tested it.  :-D  

(_And_ there was a file that can't be loaded if you don't have specific libraries.  I tweaked it to not be that way.)

adammhaile:
Yeah, I think that went in before all that.
+1

PR #423, opened 2017-09-09
Allow setting the strip_type and brightness in PiWS281X.py

rec:


bibliopixel/drivers/PiWS281X.py
```diff
@@ -42,14 +42,18 @@
     10: 0,  # Technically SPI
 }
 
+DEFAULT_STRIP_TYPE = 0x081000
```
adammhaile: We need an enum of sorts so people don't need to know the hex codes for others.

rec: I thought I looked for this before but this time I found it - these guys:  

https://github.com/jgarff/rpi_ws281x/blob/master/ws2811.h#L45-L64

Is there a canonical documentation page for these strips?  I'd love to include it in the code....

Let me know if this is right/complete and I'll make this change.

adammhaile:
Nope. At least not any in English I've ever found :P
Also, now that I look at that, we only need:
SK6812_STRIP_RGBW
WS2811_STRIP_RGB

The rest are all just channel order, which we handle ourselves. So how
about this... ditch the enum thought and just let the user choose 3 or 4
channel.

rec:
hah, I just came here to say that we handle channel order. :-)

Surely we should have a different named driver class for the SK6812?  We could have a common base class and just set 3 vs 4 channel in that class.

I really like the idea of one driver == one hardware item as much as possible - it's just so much easier to explain to people.

adammhaile:
Well... Here's the problem. You and I understand the difference between
SK6812 and WS2812 (note there are 3 and 4 channel variants of the actual
SK9812!) but other people don't understand it. Mostly I blame renaming and
branding by a certain other company that means most people know a whole
group by one name. So while different drivers make sense to us, they don't
to other people because, to them, the are all the same type of strip,
mostly.
It doesn't really help that all these chipsets are clones and ripoffs of
one other with mostly compatible protocols. Same goes for APA102 and SK9822.

rec:
> note there are 3 and 4 channel variants of the actual SK9812!

So my argument does not hold.  Gotit!  Will fix.

rec:
All right, I fixed this!

bibliopixel/drivers/PiWS281X.py
```diff
@@ -42,14 +42,21 @@
...
     """
     Driver for controlling WS281X LEDs via the rpi_ws281x C-extension.
     Only supported on the Raspberry Pi 2 & 3
     """
-    def __init__(self, num, gamma=gamma.NEOPIXEL, c_order=ChannelOrder.RGB, gpio=18,
-                 ledFreqHz=800000, ledDma=5, ledInvert=False, **kwds):
+    def __init__(
+            self, num, gamma=gamma.NEOPIXEL, c_order=ChannelOrder.RGB, gpio=18,
+            ledFreqHz=800000, ledDma=5, ledInvert=False,
+            color_channels=3, brightness=255, **kwds):
```
adammhaile: It's sad that we'll still need color_channels here when layout will be getting it eventually as well. But I had to create a custom driver for my 1-bit color display so drivers are always going to be their own thing anyways.

PR #431, opened 2017-09-16
New command `bp all_pixel_test <ledtype>` (fix #214)

### LINKED ISSUE #214: Replace AllPixel test program with `bibliopixel`.
[This page](https://github.com/ManiacalLabs/AllPixel/wiki/Library-Install-and-Strip-Example) must lose all its code.

It must be replaced by some single line that looks like:

    bibliopixel [...]

This means that we need some way to pass the LEDTYPE into `bibliopixel`.

rec:


adammhaile:
Nice! Can't it just be --numleds though? or --num?

rec:
Well, there's the shorthand "-n"....

Google and I didn't agree about everything, but they had an "almost no abbreviations" policy that I really liked.  The claim is that it's just easier to read, particularly when semi-technical people are using your end product.

But I'm really not going to make a fuss either way...

rec:
Actually, in the light of day I prefer `--number_of_pixels` as completely clear.  I can easily see hardware setups where each "pixel" was in hardware multiple LEDs.

Another clear and shorter possibility is `--pixel_count`.  We could also then use `-p` and take it away from `--path` which is mostly irrelevant now we have paths in projects (and more, is a feature that I suspect that the only person ever to use was me.  :-D)

I'm actually going to do that. brb!

adammhaile:
But then it doesn't match the API. Which has always been num. Just make it
--num
pixel_count is still too long in my opinion. I'm lazy. Also, it doesn't
match the API for both raw python and for the project files. Which is just
silly.
It needs to match the API.

rec:
Sure, _geen probleem!_  Moment pls.

rec:
Done.

PR #432, opened 2017-09-17
Simplify collection.py slightly.

rec:


rec:
Just for the record here - there is an actual change going on.

Right now, we have four ways of representing a subanimation in a collection.

1. As a string, which is the name of the animation class.
2. As a dictionary, which _is_ the animation.
3. As a dictionary, which _contains_ `"animation"` and optional `"run"` sections.
4. As a 2-tuple, `(animation, run)`

That last way is not documented, not unit tested and as far as I know no-one has used it.  I deleted it for simplicity.

(The code is really simple because cases 1 and 2 are handled centrally - we can always accept either:  `"classname"` or `{"typename": "classname"}` form.)

PR #434, opened 2017-09-17
Fix bp --help (fix #429)

### LINKED ISSUE #429: `bp --help <command>` gives wrong results;`bp <command> --help` is correct.

rec:


rec:
I have actually looked at a few other parsers in the last couple of months.  Most of them appeared better but not so much better I wanted to take the couple of hours it would take to rip out the current one and drop in a new one, plus no doubt new bugs.

I spent some time playing with the app and I wasn't able to come up with any other command line bugs, so perhaps we're through.

Certainly next time I start something, I'll look for another command line parser than `argparse`.

rec:
Hah, this literally was announced yesterday:  https://docs.python.org/3.7/library/argparse.html#argparse.ArgumentParser.parse_intermixed_args

(Basically a way to fix this and another issue I have code for in main...)

PR #461, opened 2017-10-14
Allow relative pathnames in typenames (fix #460)

### LINKED ISSUE #460: Allow relative paths in typenames.
Right now, it's a terrible drag to have to spell out `"bibliopixel.layout.matrix.Matrix"` and it's worse for animations which are going to end up having subdirectories as we end up with dozens of types of them...

Instead, each category of object (layout, animation, driver, etc) we create can have its own "base path" - for the example of layouts, it'd be `"bibliopixel.layout."` Typenames which start with `.` now mean "relative to the base path for the category of object being created.

It's really no issue to do this, as we always know what sort of category of object it is as we create it.

More, there is already code in loady that lets you can also omit the class name if it is the same, disregarding case, as the namespace!  (I must make sure that works by testing it also in BP.)

So that means that `"bibliopixel.layout.matrix.Matrix"` would reduce to `".matrix"`, which you have to admit is a huge improvement.

rec:
This came out well!  I think it'll really cut down on the cruft...

It requires loady 1.7.2 to work.


rec:
:-)

rec:
I thought of even allowing relative names _without_ the initial dot - try it without the dot, and then if it fails, try it with - but I felt it would be better to distinguish between the two cases.  Changing that decision wouldn't be hard.

adammhaile:
100% agree it should have the dot prefix

PR #469, opened 2017-10-23
Replace all naked `print` statements with `log.printer` (fix #466)

### LINKED ISSUE #466: `log` should be used for all text output
`print` should only be used for debugging!

In particular, we'll eventually want to be redirecting the output from remote bp's to your local machine, so we'll need to hijack `log`, and we can't do that for messages to `print`.

(Not of key importance but fast tasks like this I can complete while waiting for something are useful to have.)

rec:
I had a few minutes and low mental energy so did this.

Note that I had to create a new category of `log.` printer because the other ones all have some sort of preamble like `INFO` which makes them spammy.  We could also change `log.info` to have no cruft and use that, but let's wait until we actually start to redirect output and see what we need.

Honestly, if you hated this and didn't want to do it, I wouldn't care.  :-)


rec:
(Oh, I didn't change the print statements in the tests area, because we have zero need to redirect print statements in unit tests...)

rec:
Checks fail!  But I ran all the tests... :-o  Checking.

rec:
Exactly Python 3.6 failed.  And it seemed to be that same old dumb threading test.  I'll lengthen the wait somewhat in another commit.  Bounced the CI...

rec:
Now _all_ the Python 3 CI tests are failing - but I can't see a reason for it in the log...?!

Wonder if it's a Travis thing?

adammhaile:
+1 since you fixed this elsewhere

PR #474, opened 2017-10-29
flatten.py normalizes controller dictionaries

rec:
Continuing with extracting small parts of the general controllers list.

This normalizes input control dictionaries from a project.  This lets users either represent their controllers hierarchically, using addresses, or even a combination of the two.  The documentation makes it clear.

It might be a bit overengineered but I needed to normalize and check this data before I used it, and the fixup is pretty sweet.


bibliopixel/util/flatten.py
```diff
@@ -0,0 +1,44 @@
+def flatten(master):
+    """
+    Flattens a multilevel dictionary into a single-level one.
+
+        {'foo':
```
adammhaile: Needs to be a reStructruredText code block format

rec:
I pushed a tiny fixup.  Is there any way to actually see if the result is correct?

adammhaile:
Not really without parsing it with sphinx.... still working on it.

rec:
But this is right, yes?  I'll fold it and submit if so...

adammhaile:
I think so... will have to do cleanup anyways once I get it all working. So +1 for now :)

PR #477, opened 2017-11-13
Reduce number of subpackages (fix #476)

### LINKED ISSUE #476: Decrease number of second-level packages.
We have too many secondary-level packages which makes our top-level Sphinx documentation spammy.

The suggestion is to move `threads/` and `colors/` to `util/`, and `remote/`  to `animation/`.  With omitting `main/` that would give us six subdirectories, which is quite sparse for a typical Python project...

rec:
For some reason that I do not understand, the entry for `colors` in `bibliopixel/__init__.py` was _not_ enough to maintain backward compatibility - loading any class in BPA failed to work.

So I had to keep the directory with a little redirection - annoying.  

This however solves the "too busy API documents issue" as we can use whatever the command is that omits `main/` from the Sphinx documentation to omit this bogus directory `colors/`.


rec:
Oops, forgot to check the redirection part in!  :-/

Good though - it certainly proves that it was needed.

adammhaile:
Oy! Not gonna wait a few minutes to let me look at this? :P

rec:
ARG!

:-o

Sorry about that.  I was running though a lot of screens and saw the +1 _on the bug_ and...

thought it was on this.

LAME!

adammhaile:
lol... no worries.

PR #478, opened 2017-11-15
Tiny simplification to serial.py

rec:
I made a tiny change to this file while investigating imports.  If we dumped this change I wouldn't care.  :-D


rec:
(And this time I'll make sure to wait till you write +1 _for this pull request_ before checking in.  :-D)

PR #481, opened 2017-11-20
Move address_test.py to correct directory

rec:


rec:
Ignore this, my client was in an inconsistent state.

PR #482, opened 2017-11-22
Make sure that Strip.numPixels is an integer

rec:


rec:
Closed in favor of https://github.com/ManiacalLabs/BiblioPixel/pull/483

PR #495, opened 2017-11-26
v3.3.6

rec:


CHANGELIST.md
```diff
@@ -1,3 +1,6 @@
+## 3.3.6 2017-11-24
+- Fixed #485 even yet more
```
adammhaile: These are ridiculous. But not sure what else to be done.

rec:
We need better tests for this but I think I have it.

PR #496, opened 2017-11-26
`bp` can now run .py files (fix #494)

### LINKED ISSUE #494: "single class" configuration of user animation types.
People want to write a little code.  There needs to be a good "one file" mechanism for them to do that - so they only have to prepare one .py file and go.

This came up trying to explain to Harold on the user's list how to make a project.  I concluded that in his case it was more time that it was worth...!

I have a plan for this...

rec:
Think you'll like this.

Now you can write a single .py file with an animation in it and run it directly, if that file has a static member called `PROJECT` - or the animation class has a static member called `PROJECT`.

It uses the loady heuristics to guess which symbol in the file you want for your animation, or if that fails, looks for a symbol `Animation` - so basically if you name your class "the same name as the file" or "Animation" it all works automatically.

The example should make it clearer.

The motivation was to give Harold a quick way to port his tiny demo program to use projects.


rec:
The change removing the old "projects" was broken, and so is this now due to rebasing issues.

`dev` is fine but I'm going to have to retweak these two changes.

adammhaile:
+1

rec:
Just did all the rebasing and waiting for the final CI tests to work.  This will be really nice, and once we have this pushed to master, I'm going to emit a document on "How to write an animation" which should let people just write a single Python file and go...

PR #497, opened 2017-11-27
Remove old project code (the sequel)

rec:
Now with new improved Correctness™!  (The previous try broke the ability to run projects which contained nested animations like Sequence, Remote, etc...)


bibliopixel/animation/runner.py
```diff
@@ -4,7 +4,7 @@
 
 class Runner(object):
 
-    def __init__(self, amt=1, fps=None, sleep_time=0, max_steps=0,
+    def __init__(self, *, amt=1, fps=None, sleep_time=0, max_steps=0,
```
adammhaile: WTF is that? Please at the very least comment so people who aren't you and Guido know what on earth that asterisk does and means.

rec: The `*` means "this function takes no positional arguments".  From [here](https://docs.python.org/3.2/reference/compound_stmts.html#function-definitions), "Parameters after “*” or “*identifier” are keyword-only parameters and may only be passed used keyword arguments."

`Runner` is only intended to be constructed like a dictionary - `Runner(fps=30, max_steps=1024, until_complete=True)` and never with positional arguments - `Runner(1, 30, 0, 1024, True)` - and this disallows this second form.  It's just good practice and more, I ran into `amt` being a miscaptured parameter in debugging, so I put it in to find the original place where it happened.

This is standard Python 3, so I'm really not sure if it needs to be documented here.  If someone is really reading to that point, they should know this or at least be willing to look it up.  :-D

The code will work without it too - I'd rather delete it than document it.  :-D

TBH, I only learned about that last year...

adammhaile:
Leave it. I just wanted it on record somewhere. I realize it's standard
Python 3 but much of Py3 is still kinda a mystery to a lot of people ;)
+1 on all this.

Issue #223, opened 2017-06-16
Use documentation autogenerated from the Python code in place of the hard-coded documentation in the Wiki.


rec:
If you look at many of the pages on the Wiki, they contain documentation of code:  e.g. https://github.com/ManiacalLabs/BiblioPixel/wiki/Matrix

Unfortunately, we edit the code and the documentation doesn't keep up.  Also, the formatting isn't so good but we have no way to change it en masse. 

We should be using Sphinx to generate all the documentation right from the Python files.  We should be - somehow - pushing that documentation to the Wiki, or linking to that documentation in the Wiki.  We should be deleting all specific documentation of methods from the Wiki, to avoid having to change everything in two places, and always linking to the generated docs as the One Source Of Truth.

rec:
This has some more information: http://martin.majlis.cz/integrating-github-with-pypi-travis-ci-readthedocs-and-code-climate/

I'm also agreeing that we might want to push to ReadTheDocs.com.

PR #498, opened 2017-11-28
Sphinx

adammhaile:
Ok... I've been messing with this forever but at this point I think, although it's a raging mess, it needs to just make it into dev so we can have more than my eyes on it. I also need opinions as to direction...

So, first a primer:

Install sphinx:
`sudo pip install ./doc/doc_requirements.txt`

Generate docs:
`./gen_docs.sh`

There will be a ton of warnings. This is expected for now.

Then open `./doc/_build/index.html` in your browser.

As you'll see, I have a basic home page with a simple TOCOk... I've been messing with this forever but at this point I think, although it's a raging mess, it needs to just make it into dev so we can have more than my eyes on it. I also need opinions as to direction...

So, first a primer:

Install sphinx:
`sudo pip install ./doc/doc_requirements.txt`

Generate docs:
`./gen_docs.sh`

There will be a ton of warnings. This is expected for now.

Then open `./doc/_build/index.html` in your browser.

As you'll see, I have a basic home page with a simple document TOC and an API TOC. @rec , you'll likely be most interested in the API TOC to see everything that is auto-generated.

So, here's the open issues:

- Most *.py files have terrible docstrings. Obviously this needs to be fixed and is part of the reason I'm PRing this in an incomplete state. There's a lot of work to be done there and having this in dev will benefit that as we can verify those new docstrings as we go.
- There is a way to do doc coverage reporting. I will of course add this but it's moot now.
- The biggest question here is what of the old docs, if anything to keep. Many of them can become just straight up API docs, though as you'll note, the layout of those is not clear. I think we should have the basic API TOC, but maybe also call out some specific stuff in the main TOC... or maybe none at all?? I mean I guess it should go all in on projects. Anyways, I would love your opinion on the bare minimum list for non API doc pages we need here and what can go.


rec:
This is going to make a huge difference in how usable this program is for other people - this and a few tutorials and we'll actually be able to lure in third parties.

A huge amount of work here, and non-programming too - my hat is off to you!

1. This is against master!  Could you reopen this pull request against dev, please?
 http://sadtrombone.com  

2. I suggest the commit message `Automatically generate documentation using Sphinx (fix #223)`

3. [lots of blahblah deleted here].

I can go through my long reasoning process if you like, but checking in generated files is a mortal sin, particularly in this case.  :-D

I'm going to **strongly** suggest you not check in the generated documentation files and _just_ use https://readthedocs.org instead.  They do ["continuous documentation"](https://docs.readthedocs.io/en/latest/index.html) and thus there's no need to check in anything at all. It'll mean that we'll always have up-to-date documentation trees for `dev` and `master` without us having to do anything at all.

You might not even need the .sh and .bat files - it might be possible that https://readthedocs.org can even generate documentation for branches under review like Travis does.

rec:
So it seems to do all we need - plus it automatically generates PDFs (which aren't essential but nice-to-have) and man pages if set up nicely.  A man page for the `bp` command would be awesome...

adammhaile:
- No longer against master. You can change that after the fact! Who knew?
- Commit message changed
- **Nothing** generated is part of this!! You'll note that `.gitignore` has added `doc/api` to the list which is the only thing auto generated by this. The rest of the files in the `doc` folder are just converted from the wiki markdown files. I fully expect some (a lot) of these to go away. But nothing committed is generated. I would never do that. Everything here is needed.
- ReadTheDocs is coming and was always planned. I have the account setup for Maniacal Labs. I just wanted to not push to it until we had docs that weren't a complete mess :P
- `man` page would be cool but no idea how that's done. Also that would be completely separate from this as right now `bp` is not documented in any automated way. I'll look into some way of adding that, but not in the scope of this.
- PDFs are easy. Just adding a flag to the output options. Again... that's once things are more stable.

rec:
Ah, now I can see the complete diff it's easier.  You basically just ported over all the .rsts from the Wiki, gottit.

Alright, this is good to go.  The man pages aspect of ReadTheDocs is just some setting, I'll put that in when it's all running and lovely.

:+1:

PR #499, opened 2017-11-29
extractor.py extracts fields from incoming messages.

rec:
More material from the controller branch.

Again, like `Address` I hate to check things in without an immediate use but there is simply too much stuff, so I have to check in building blocks which aren't yet used.

This won't affect the general public or correctness of existing code so it's not a risk.

This `Extractor` class is motivated by MIDI as you can see by the test, but it's perfectly general.  I think the only obscure idea is `auto_omit` but you can just leave that alone...


rec:
Travis gives me consistent, inexplicable errors, but only on Python 3.6 - and I have a clean 3.6 virtualenv which does not show this issue.

https://travis-ci.org/rec/BiblioPixel/jobs/308941245: `ERROR: file not found: discover`

rec:
I'm going to ignore that issue and push it anyway.  I see it in other BP branches of mine :-/ which have nothing to do with this, so I think it's some systemic issue.

PR #500, opened 2017-11-29
 Generator and Midi.

rec:


rec:
Ach, Travis, you're getting me down.  :-/

---

I renamed everything and added more documentation.  Particularly, `Generator` was a terrible name.  `ControlSource` says everything.  :-)

rec:
And let me know if you have more questions about this idea.  I think it's pretty straight-forward, but if that's the case I should be able to explain it really well.

rec:
Also - I am working on a further changes in this design to deal with threaded vs. un-threaded (some sources like the keyboard driver I picked come with their own threads intrinsically, others like Midi do not).

It builds logically on this commit, but if you liked, I could wait and rebase them together to see the final result.

rec:
I'm closing this, and re-opening another one with the final version of this code.

PR #503, opened 2017-11-30
fixup! Pass test command line around in setup.py (fix #442)

### LINKED ISSUE #442: No way to run individual test suites or modules.
No matter what, setup.py just runs all the tests - it seems to ignore its arguments:

    python setup.py test --test-module=test.bibliopixel.project.project 
    python setup.py test --test-suite=test.bibliopixel.project.project.Project
    python setup.py test --test-module XXX.XXX

all do exactly the same things.

rec:
For whatever reason, under 3.6 we're now getting passed `setup.py test discover` where we weren't before.

This brutal change fixes the problem, though it means you can't pass flags through setup.py to pytest any more.

This feature has always been problematic, however, as most of the useful flags don't get passed through anyway.

Instead, if I need to pass arguments to pytest, I call it directly.  Less typing anyway.

Note this is a fixup.  With your permission, I'm going to rebase it into dev to clean up all these failed pulls and commits.


rec:
Fixes #501

adammhaile:
lgtm

rec:
Should I fold it into the original commit on dev?  That was my plan - otherwise, it means that any commit between the original commit and now will seem broken.

I wish Travis didn't change underneath us - how is this happening?

adammhaile:
Yeah, that would probably be best.

rec:
The rebase is complete.

PR #506, opened 2017-12-04
Final project refactoring.

rec:
Ah, so sorry that there's more of this stuff, but it fixes some bugs that came up and also is much easier now. 

Let me know what I can do to make this easier to read for you...


rec:
Let me summarize also why this was necessary, and why we are better off.

There is a known (to me) known bug in the code [here](https://github.com/ManiacalLabs/BiblioPixel/blob/dev/bibliopixel/project/project.py#L21-L22) involving constructing parts of an animation.

You see that line only constructs things that take a `layout` as their first parameters.  But a lot of things we need to put inside an animation don't care about `layout` one way or the other.

In general, there was a lot of code there just dealing with the different construction parameters of classes within a project, and it was a source of problems during development.

I realized that all objects in a project (animations, layouts, devices and soon control sources and more) could have a uniform constructor - with exactly one positional argument,the project, and then named arguments as before.

Now, we can't change constructors within a version.  

But this idea if we could use it cleans up a ton of crap.  The animation, layout, driver or whatever just takes what it needs from the project.  We can add all sorts of new facilities to the project, and animations, layouts and any new classes we create in future can get them right from the project without changing any API.    

Well, I figured out a way to make that work without changing any existing code.  I just added [a new optional classmethod](https://github.com/rec/BiblioPixel/blob/project/bibliopixel/layout/layout.py#L9-L13) on a few classes which is the "new" constructor.  During construction, [we try to use that static function if it exists, otherwise fall back on the classic constructor.](https://github.com/rec/BiblioPixel/blob/project/bibliopixel/project/project.py#L53-L57)

It's a tiny amount of code and it fixes all of the issues above.  If we feel like it, in 4.0 or whatever we can change all our constructors to look like `__init__(self, project, ...)` and get rid of the bit extra, just for neatness.

Doing this also let me get rid of the `RunAnimation` abomination, and also get rid of the per-class code for emitting children, resulting in two fewer .py files and about twenty lines less of code, as well as additional functionality.

adammhaile:
Nice. And I'm all for next release being 4.0, so change the constructors to your heart's content.

rec:
I think we need a maintenance last 3.x release.  It'll bring particularly "running .py files from `bp`" which I think is a really useful feature...

adammhaile:
Ok... that's fair. Let's then do a 3.5 which will be maintenance, and then
go hog wild on 4.0

PR #507, opened 2017-12-05
Per-project aliases

rec:
* New `aliases` section in a project
* Example [here](https://github.com/rec/BiblioPixel/blob/aliases/test/bibliopixel/project/project_test.py#L245-L263)
* So useful when I started on my own Christmas lights project
* If you exclude the new test, there are 14 lines code _less_ and more functionality


rec:
Ping!  A nice little feature that makes big documents much clearer.

PR #509, opened 2017-12-06
Allow strings as colors in `layout.set` (fix #508)

### LINKED ISSUE #508: Allow strings as colors in `layout.set`
We should allow `layout.set` and the various methods on derived classes to take a string as a pixel value, which we automatically convert to an RGB value (or throw an exception).

rec:
There's no unit test that covers this.  Suggestions?


adammhaile:
lgtm

rec:
(I haven't forgotten this - I just need to figure out a unit test...)

rec:
All right, I wrote a unit test - ready to go.

adammhaile:
+1

PR #510, opened 2017-12-06
Allow animations in projects to be classes.

rec:
* Really useful for ".py projects"


adammhaile:
Umm... couldn't it always do this? I mean, I've been doing BiblioPixelAnimations.matrix.MyAnimationClass forever in projects.

rec:
No, you've been using the string `'BiblioPixelAnimations.matrix.MyAnimationClass'`, not the actual class.  :-D

This is for ".py project files" which I don't think you've used - see [this line](https://github.com/ManiacalLabs/BiblioPixel/pull/510/files#diff-8fb07d630ca31b5d85059b7e03d96365R21) to see the difference.  You can't even make this happen in JSON.

adammhaile:
Oh... gotcha. Cool. +1

PR #513, opened 2017-12-08
`dimensions` section in project file.

rec:
Another thing to make projects easier.

You can now set a "dimensions" section in the project file or the `--dims` flag command line and the program works out the right `num`, `layout`, and parameters to the layout (`width` and `height` or `x`, `y`, `z`.

This means that nearly all the time you don't need to specify a layout section at all... :-)


bibliopixel/project/fix.py
```diff
@@ -0,0 +1,92 @@
+import copy
```
adammhaile: Why is this called fix?

rec: Uh, good question.  :-)

The idea is that they "fix" up a project description.

Not a great name izzit?  Ideas?

adammhaile:
Nope... I hate it. It's ambiguous. Even cleanup would be better.

rec:
`cleanup` is better but there must be something better than that.

Let me think about this a bit more, then... :-)

rec:
I decided `cleanup` was as good as anything and I made the change. What do you think?

adammhaile:
Works for me.

PR #515, opened 2017-12-11
User defaults for projects

rec:
* New `bp default` command

I wanted to get this in before 3.5.

If you recall, I had a much less nice version of this last year sometime.

This hits the spot - which means that people should be able to run each other's project without changing a thing if they set a couple of defaults...


rec:
(A later exercise might be to prompt the user and ask them questions about their drivers and layout, and then make a defaults file.  Now we have this, it should be easy...)

rec:
Eh, I can't request a pull request on a branch starting with x- - my fault!  I'll close and re-open this.

rec:
(Or rather, I can do that, but it didn't report my failing branch...)

PR #519, opened 2017-12-11
Add a new logging level, log.frame

rec:
* Currently, log.debug is hard to use because it prints one line per frame


rec:
I'm discovering all this stuff when I'm trying to develop my own "third party" animations as if I don't know the program.  Very instructive.

PR #522, opened 2017-12-13
Improve error if numpy is requested but missing

rec:


bibliopixel/main/common_flags.py
```diff
@@ -102,7 +102,7 @@ def add_project_flags(parser):
         help='Number of pixels controlled by the driver.')
 
     parser.add_argument(
-        '--numpy_dtype', default=None, help='Use this numpy datatype.')
+        '--numpy_dtype', '-m', default=None, help='Use this numpy datatype.')
```
adammhaile: -m ? not -n?

rec:
We use `-n` already for "number of pixels".

Actually, now we have `--dimensions`/`--dim`, that sort of takes the place of `-n`, so I could obsolete it...

adammhaile:
Yeah, that's better.

rec:
Done!

rec:
Ah, I pushed an extra commit onto the end of this sequence which makes life much simpler for the user - no more obscure "numpy_dtype".

Now you can say things like:

    bp lites.json
    bp lites.json --numbers=python
    bp lites.json -n python  
    # Those last three are all the same.

    bp lites.json -n int
    bp lites.json -n float
    bp lites.json -n uint8

or in your project:

    {"numbers": "float"},

or in your animation class:

    class MyAnimation(BaseAnimation):
        PROJECT = {"numbers": "float"}

adammhaile:
Very cool. +1

PR #524, opened 2017-12-15
Slideshow mode for sequences.

rec:
* --slideshow flag lets you set this value in the top-most sequence

* Also, child animations without fps or sleep_time settings now get those
  values from their parent


rec:
Yes, this is super-useful for my Christmas lights testing.  :-)

PR #527, opened 2017-12-16
Better errors if `serial` is installed, not `pyserial` (fix #526)

### LINKED ISSUE #526: Better error if someone accidentally installs `serial` instead of `pyserial`.

rec:


rec:
I experimented with this, and strangely, you don't need the uninstall if you install `pyserial` over `serial`, but you _do_ need if you were for some strange reason installing `serial` over `pyserial`.

I left the uninstall in anyway, because it can't hurt, and there might be some configuration where it works better that way.

rec:
Yeah, neither did I until I did it by mistake on the Rpi.

PR #534, opened 2017-12-16
Allow "numbers" section to have a default

rec:
Oops.  

Everything else in the previous commit works except setting a default value for numbers.

Once this is checked in, I urge you to `bp defaults set numbers=float` for a while to see how compatible the two modes are!


rec:
Ah, I think I rebased those changes in badly.  Stand by...

rec:
Ping!  Tiny one...

adammhaile:
+1

PR #535, opened 2017-12-16
More commands relating to defaults.

rec:
I am finally putting back in the loading and saving defaults commands.

This is SO useful for me - I have two defaults files, one for simpixel and one for my strips (where I had to tweak the gamma correction), so now I can just run any file with 

    b proj.json -d simpixel

or

    b proj.json -d strip-gamma

and just switch back and forth that way without editing the files!


rec:
Ah, I think I rebased against dev wrong.  Stand by...

rec:
Ah, that's better...

adammhaile:
+1

PR #537, opened 2017-12-18
Fix #536

### LINKED ISSUE #536: Remote is broken in dev
This is 100% my fault: my standard test simply didn't include `Remote` because its test didn't start with `0`.  :-o

However, porting `Remote` is going to be tricky, so let me write out the issues.

The big problem with `Remote` is that it has an extremely long and complex constructor that relies on the fact that its arguments have not yet been constructed.

Big constructors are intrinsically bad for numerous reasons (hard to refactor, makes an empty constructor impossible, hard to test, ties a lot of key code into a single knot, etc).

In this case, we have the issue that in the new world, Animations are passed all the subelements completely constructed.

(Which completely makes sense.  You as the "new class writer" shouldn't have to know how to construct all the other sorts of things you get - the framework should handle that.)

In fact, the majority of `Remote`'s constructor consists of rewriting the dictionary that describes its parts.

We have a mechanism for that, now - there's a static class function:

    @staticmethod
    def pre_recursion(desc):

that rewrites the dictionary if necessary before the constructor is called.

--

Digression:

Again, it worries me that we have assumed a level of complication like this, but I feel this is justified, because our real-world objects are intrinsically fairly complicated.

In this case, we have a description of a project and all its parts as a human-friendly dictionary, and we need to rewrite that into a "canonical" dictionary that we can use for constructing the actual objects in our system.

The whole thing is complicated - there are a huge number of possibilities and also many places we make things easier for the user.  I made the split because of the difficulty in keeping all the construction and error reporting straight.

So I believe dividing this into two phases - project rewriting and construction - makes things _easier_ in the long term.  

So it makes sense to have this two-phase construction - and you can see that the code in remote parallels this idea with its rewriting.

--

OK - so I basically move all the code that rewrites the dictionary out of the constructor into the project rewriting phase.

I'll find when I made the error - it might have been a while ago, but it's still in dev, so no one else has it -- and plan it as a --fixup for that.

--

I see this as the system working well (given my mistake in not checking remote).

If I hadn't caught this, you would have, when we made the release to master.  dev should be known to be less reliable and potentially subject to rebase.

We'll rebase this in, present a coherent and correct 3.4, and I'll fix my procedure to run remote as a separate step.  (I need something that pops up the remote URL as part of the process, I'll write that too...)

(Or if rebasing is too onerous, having a separate commit later would be fine in this case - it's not like it's breaking any tests...)

rec:


bibliopixel/animation/remote/control.py
```diff
@@ -172,7 +179,10 @@ def run_animation(self, name=None):
             name = self.default
 
         if name not in self.name_map:
-            return False, 'Invalid animation name: {}'.format(name)
+            error = 'Invalid animation name: {}'.format(name)
+            log.info(error)
+            print('!!!!!', self.name_map)
```
adammhaile: Left over debug print? We shouldn't have *any* print() statements in the code. Just log.<something>

rec: Urg!  :-D 

Pushed a fix.

rec:
Fixed!

PR #548, opened 2017-12-21
Add color_list setters in Layout and Animation

rec:
I keep writing this in client code:

    self.color_list[:] = compute_something()

but that doesn't work:  I have to write

    cl = self.color_list
    cl{:] = compute_something()

which is really stupid.  

The setters fix this.


adammhaile:
Weird... wonder why?
+1

rec:
Ah, there is a reason for this - it's because `color_list` is a property, so when I do any sort of assignment, even `layout.color_list[:] = ...`, it tries to call the setter for that property - which doesn't exist.

PR #549, opened 2017-12-21
Better error reporting of project errors

rec:
Results in errors list this:

    ValueError: Project section "animation" is list, should be dictionary


adammhaile:
Very nice :) +1

PR #557, opened 2018-01-01
Fix `bp demo`

rec:
There's a weirdness with the circle demo that does something more interesting than is actually programmed in :-o but it'll work fine as a demo.


adammhaile:
Wait... circle wasn't causing the issue though, right? What's wrong with Circle?

rec:
All five demos failed, at least when I tried them!  One was a dupe...

adammhaile:
I know all five failed. That's why I was asking what was specific about circle.

rec:
Well, I changed circle to be a project, because then I could get rid of that whole codepath where demos weren't projects...

I'm not exactly sure what it was doing before, to be honest.

adammhaile:
lol
+1

PR #562, opened 2018-01-07
Rename `slideshow` to `length`

rec:


rec:
Nah, I mentioned doing it in the future.  You must be confusing the future for the past again.  :-D

PR #567, opened 2018-01-09
Add pyserial as a dependency (fix #553)

### LINKED ISSUE #553: `pyserial` should be a dependency.
We are already installing so much stuff, one more thing won't hurt, and this will prevent anyone from installing the package `serial` instead of `pyserial`.

rec:


adammhaile:
+1

PR #570, opened 2018-01-12
Fix SPI driver

rec:
Not sure how this ever worked!


adammhaile:
Confirmed works. Rebasing in right away.

PR #572, opened 2018-01-12
Allow Collection to load animations from files

rec:
A bunch of commits from last month were with the goal of this change, which makes combining existing animations into one sequence (or other collection) a piece of cake.


rec:
Ping!

adammhaile:
Oops. Thought I already +1'd this. Looks good :)

rec:
This is my favorite new feature.  When you start doing big projects with lots of new animations, you'll see what I mean.

PR #581, opened 2018-01-17
Re-enable log.info (fix #580)

### LINKED ISSUE #580: No message output for Serial reboot with --loglevel frame
```
G:\>bp run test.json --loglevel frame
DEBUG - run - Running file test.json 
DEBUG - animation_threading - Animation starts on main thread
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
FRAME - animation - 0ms/1000fps / Frame: 0ms / Update: 0ms
```

test.json
```
{
  "animation": {
    "typename": "strip_test"
  },
  "driver": {
    "num": 12,
    "typename": "serial",
    "c_order": "RGB",
    "ledtype": "APA102"
  },
  "layout": {
    "typename": "strip",
    "brightness": 255,
    "threadedUpdate": true
  }
}
```

It should show a message about having to reconfigure and reboot the allpixel. But that's lost.
Looks like log.info() is just not working: https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/drivers/serial/driver.py#L48

rec:
I shouldn't have to add each of these individually - we didn't before - but the only way to get consistent results appears to be doing that.  As long as it works!


adammhaile:
Wow... Seriously? I hate python logging.
Looks good.

rec:
Just FYI, I tested all 25 combinations of loglevel and logging statement before checking this in, so it had better work.

PR #589, opened 2018-01-23
Unflatten unpacks flat arrays into deep ones

rec:
Going back to the general controllers branch!

 If you recall, I need "flatten" to represent documents, and it turns out that I also need an "unflatten".


rec:
Closed in favor of #593.

PR #590, opened 2018-01-25
BaseAnimation.step() now does nothing

rec:
* It's perfectly reasonable for an animation to do nothing in `step` -
  for example, if it's a static animation that does all its work in `pre_run`


rec:
Closed in favor of #593.

PR #591, opened 2018-01-26
Better text for `bp --help`

rec:
* Fixed `bp -h` to give the same results as `bp --help`


adammhaile:
+1

rec:
The help is actually pretty useful at this point, shockingly enough.

PR #596, opened 2018-01-30
Make float32 the default float type for numpy

rec:
There's really no advantage to `float64` for colors and it doubles our memory footprint.  You can always explicitly ask for `--numbers=float64` if you want them.


adammhaile:
+1 Good idea

rec:
I wrote some more Project documentation on the Wiki, too.

adammhaile:
Awesome. But don't put too much more in there... what I've got as sphinx
docs is just a snapshot and we need to not loose stuff as we start doing
the transition. I started working on adding docstrings last night.

rec:
Ah, yes, I was aware of that transition but I had a few minutes in this hotel room and wanted to do a quick edit.  Also, I'd just sent the link to the list and saw it was out of date.

When I get back, let's figure out exactly how to go forward with docs in future...

PR #597, opened 2018-01-31
[WIP] Beginning examples of apidoc setup

adammhaile:
Just getting started and hitting a few points to show what needs to be done. See comments in diffs.


bibliopixel/util/threads/compose_events.py
```diff
@@ -2,10 +2,12 @@
 
 
 def compose_events(events, condition=all):
-    """Compose a sequence of events into one event.
-    Arguments:
-        events:    a sequence of objects looking like threading.Event
-        condition: a function taking a sequence of bools and returning a bool.
+    """compose_events(events, condition=all)
```
adammhaile: ALL docstrings are to define function params as shown here. Any that exist with the old format should be changed. All new ones must follow this pattern. See here for more: http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

example:
```
:param str sender: The person sending the message
   :param str recipient: The recipient of the message
   :param str message_body: The body of the message
   :param priority: The priority of the message, can be a number 1-5
   :type priority: integer or None
   :return: the message id
   :rtype: int
   :raises ValueError: if the message_body exceeds 160 characters
   :raises TypeError: if the message_body is not a basestring
```

bibliopixel/util/threads/sub.py
```diff
@@ -14,18 +14,15 @@ def run(function, *args, use_subprocess=False, daemon=True, **kwds):
...
-        use_subprocess: if true, create a new multiprocess;  if false, create
-            a new thread
-        function: the function to call
-        *args, **kwds: arguments to the function.
-        daemon: is the thread or subprocess run as a daemon or not?
-
-    Returns:
-        a tuple with three elements: the subprocess or thread, an input
-        queue, and an output queue.
+        function(input, output, \*args, \*\*kwds)
```
adammhaile: * is special in reST, must be escaped.

bibliopixel/util/threads/sub.py
```diff
@@ -14,18 +14,15 @@ def run(function, *args, use_subprocess=False, daemon=True, **kwds):
...
+
+    :param use_subprocess: if true, create a new multiprocess;  if false, create a new thread
```
adammhaile: ALL docstrings are to define function params as shown here. Any that exist with the old format should be changed. All new ones must follow this pattern. See here for more: http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

example:
```
:param str sender: The person sending the message
   :param str recipient: The recipient of the message
   :param str message_body: The body of the message
   :param priority: The priority of the message, can be a number 1-5
   :type priority: integer or None
   :return: the message id
   :rtype: int
   :raises ValueError: if the message_body exceeds 160 characters
   :raises TypeError: if the message_body is not a basestring
```

bibliopixel/util/threads/compose_events.py
rec: I don't quite understand this first half of the comment.  More, shouldn't `all` be in quotes?

bibliopixel/util/threads/sub.py
rec: Is there a way to automatically identify all functions that _do not_ fit this criterion?

(My guess - probably not/not worth doing ourselves...)

bibliopixel/util/threads/compose_events.py
rec: Also, we lose information in this edit and in the next one too...!

rec:
All right, I have understood all this and have a couple of sphinx pages up in parallel tabs ready to start doing this...

bibliopixel/util/threads/compose_events.py
adammhaile: No... because all is literally the python built-in function`all` - If you look in the code, it's not a string. It's just that function object.
And yeah... I removed the return value details. Will add back.

bibliopixel/util/threads/sub.py
adammhaile: Not really. Aside from just grep :P We *need* to go through every file in the whole project though and cleanup/add/update the docstrings, so we'll find them all.

adammhaile:
@rec - Cool. We need to coordinate on what files each of us are working in. To be honest, there's a lot at this point that you're going to be the better source of what's what. But I'll work on it where I can. So in general, if you start on a section, just let me know and I'll stay out of that part. And I"ll do the same.

Also... one note: We have *a lot* of places where the base class has a ton of parameters but all the classes that inherit from that base just do `**kwds`. We can either copy/paste the same list of valid params to each instance, or (and this is what I prefer) reference the base class in the inherited class docstring to say something like `Matrix includes all the shown parameters above as well as all those allowed in :py:class:'bibliopixel.layout.Layout'`
That `:py:class:` stuff should automatically make a callout and link to the right place.
Please be sure to use the gen_docs.sh script to render and confirm everything looks as expected.

rec:
I like the "includes all the shown parameters" wording too.

I opened a new issue, #605, for us to work on this whole problem.

Feel free to submit this!

rec:
Er, once you've put back the removed text somewhere, even if only as a comment.

bibliopixel/util/threads/compose_events.py
adammhaile: Wait... what is getting lost here? I see nothing that didn't make it into the new stuff.

PR #608, opened 2018-02-03
More docs

rec:
Colors.

You'll note that I just started on `util/thread` and then realized that there was too much stuff that no one cares about and is going away in 4.0 anyway.  We need a way to mark some directories as "not documented".


bibliopixel/util/colors/arithmetic.py
```diff
@@ -1,10 +1,16 @@
 def color_blend(a, b):
-    """Performs a Screen blend on RGB color tuples, a and b"""
+    """
+    Performs a Screen blend on RGB color tuples, a and b
+    DEPRECATED:  use numpy
```
adammhaile: Show example or provide link. This is useless to someone who doesn't know numpy.

bibliopixel/util/colors/arithmetic.py
```diff
@@ -1,10 +1,16 @@
...
     return (255 - (((255 - a[0]) * (255 - b[0])) >> 8),
             255 - (((255 - a[1]) * (255 - b[1])) >> 8),
             255 - (((255 - a[2]) * (255 - b[2])) >> 8))
 
 
 def color_scale(color, level):
-    """Scale RGB tuple by level, 0 - 256"""
+    """
+    Scale RGB tuple by level, 0 - 256
+    DEPRECATED:  use numpy
```
adammhaile: Same as above.

bibliopixel/util/colors/gamma.py
```diff
@@ -1,5 +1,17 @@
 class Gamma(object):
+    """
+    Compute a fixed gamma table with 256 entries.
+
+    DEPRECATED: use numpy
```
adammhaile: Same as above.

bibliopixel/util/colors/arithmetic.py
rec: Actually - I'm going to eliminate these files from the .rst build entirely!

adammhaile:
Fair

rec:
All right, I tweaked it a bit more!

The docs look OK.  I can see a bunch of ways to make 'em better, but I want to get through getting at least superficial coverage on everything before going back and polishing.

rec:
Are we good to go?  :-D

adammhaile:
Yep, looks good.

PR #611, opened 2018-02-05
Add fail parameter to serial.Devices.error

rec:
We actually use that parameter elsewhere.  :-o


rec:
Should we push a new release?  This is to fix someone's bug...

adammhaile:
Sure.

PR #616, opened 2018-02-10
Fix crash in `bp demo`

rec:


rec:
Wilco!

The root issue is that `bp demo` is not automatically tested.

PR #621, opened 2018-02-17
Catch and report all exceptions from serial communication

rec:
I haven't had a chance to run this on my own strips, but they won't really test this anyway, since I don't get these errors.

Instead, I'm going to try to get the user reporting the issue to patch this commit in and see it if works better for them.  Stand by!


bibliopixel/drivers/serial/driver.py
```diff
@@ -144,7 +146,14 @@ def _compute_packet(self):
         self._packet.extend([0] * self._bufPad)
 
     def _send_sync(self):
-        self._com.write(self._sync_packet)
+        self._serial('write', self._sync_packet)
+
+    def _serial(self, method, *args):
```
adammhaile: This is going to slow things down because it's got to do these weird abstractions every time. We need serial to be as fast as possible.

rec:
I think it's a false optimization - the real cost of that phase is the endless list manipulation to render a buffer! - but I did it anyway.  :-D

adammhaile:
Thank you. We gotta keep absolutely minimum of crap in the way of writing out to serial.

rec:
I'll submit this when I see that the user with this issue can run without crashing... :-)

rec:
So I'm going to push this unless there are any objections.

I had an idea, though.  Is there a way that you could come up with to reliably trigger serial errors?  

I'm thinking that if you just had a decent bench power supply, which you must have, you could start to reduce the voltage or restrict the current in some setup until it started to fail.

adammhaile:
Been there. Tried that. In the history of the allpixel, which has sold
thousands of units, I've seen this come up maybe 4 times. When reducing the
voltage it just eventually cuts out for good, not intermittent.

PR #623, opened 2018-02-17
New flag `bp --ignore_exceptions` (fix #622)

### LINKED ISSUE #622: If you are running multiple documents from `bp` then control-C only interrupts the current document
I've known this for a long time and how to fix but keep forgetting to do so.

rec:


rec:
This was the default behavior before and it never bothered you!  :-D

We don't ignore _all_ exceptions - from the commit:  

    'If true, continue running the next project if one project fails'

This is only applicable if you are running a series of projects like `bp foo.json bar.json`. 

As it is now, if you fail on one (either through a keyboard interrupt, or some bug), `bp` continues to run any remaining projects.  After the change, that behavior only happens if you _enable_ this flag - by default, it will throw an exception and terminate the whole series on the _first_ exception.

adammhaile:
Ah, sorry... misread that. Yeah, this is fine.

rec:
It wasn't so obvious. :-)

PR #629, opened 2018-02-19
Simplify animation code and fix #169

### LINKED ISSUE #169: A "step" in `Sequence` can take arbitrarily long.
Because `Sequence.step` calls `run()` for each animation it contains, if one of those animations goes on forever, the whole sequence goes on forever.

This is bad for re-use.  It means I have to make many copies of one subsequence, each having a specific length.  I'd rather make one sequence of infinite length, and then when I re-use it, just run it for a specific time. 

Instead of calling `run()` the sequence needs to call `step()` on its subsequence as many times as it needs to.

In order to do this, the idea of `step` needs to be firmed up somewhat.

rec:
#169 

I need "one sequence step == one subanimation step" for a host of things, and one branch that I'm almost finished with.

This is very delicate code, so I was extremely careful to try and make it as clear as possible, even if it involved more commits overall.

The first three commits are pure refactorings to make the last commit - which changes the behavior of `Sequence.step()` to fix #169 - both as short and clear as possible.

I think the code ended up better for the refactorings - at least, I understood it better.

I will have another more exciting review "fairly soon" with that other branch.


rec:
Ah, this is a real breakage, but why didn't I see it in client?

I'll fix it, one moment.

rec:
No, it isn't a real breakage!  

Github has this commit ID: d82012
So do I - that's the one I tested.

But Travis has d9d9024!

(That was a branch I had earlier today under that name.  It's long force-pushed over...)

I'll bounce Travis and try again.

It seems to be that it is testing a version of the branch the way it was an hour ago, which is why I thought it was a real error.  "Didn't I fix that?"  :-D

adammhaile:
This is really getting into big change territory... I thought we were keeping anything big out of 3.5 doc release?

rec:
Ah, well, this forces me to show my hand.  :-D

I did this because I have written an Animation(*), a Wrapper that writes animated GIFs (while also passing the information through to your driver).

gif_writer

I feel we need this for three big reasons:

1. In documentation, I want to be able to say, "If you run this project and it will look like _this_.  Change this, and it looks like _that_."  Lots of little pictures will make the documentation far more readable and more fun to write.

2. When we announce new features, I want to be able to send people a taste of those features.

3. I want (eventually) to write a huge regression test that runs "all" our animations and makes sure that they are identical.

However, for this to work, the containing animation needs to get called back each frame - not "at the end of each subanimation".  I only realized that after I had finished the GIF animator.  Thus, this review...

I would note that it doesn't change any visible functionality at all _unless_ you are a developer who has written a class that derives from `Sequence` - but if you, you were likely one disappointed person...

--

(* - Yes, I also thought it might have been better as a driver - but it needs to know about the layout, and drivers don't know about layouts, by your design, a decision I strongly agree with.)

adammhaile:
LOL... I've written the GIF output too. Just don't think I ever commited
it. Though mine was a driver. Yours is probably better.
Alright, carry on. +1

PR #630, opened 2018-02-20
Gif snapshots

rec:
All right, the big one.  :-D

As usual, all but the last are pretty trivial.  I did find a problem with collection.Wrapper that I fixed.

Writing GIF animation is a bit of an art, and there's an issue in PIL that prevents you from writing animations with more than about 200 or so frames.  

I have code that should fix it but runs into other issues in PIL, which has some funkiness around it.  

Or there's a wrapper around PIL called `imageio` that lets you circumvent that issue but has [other issues](https://github.com/imageio/imageio/issues/312) (and I found the time durations were screwed up too but I haven't reported that yet...)


rec:
Ping!!

adammhaile:
+1

PR #634, opened 2018-02-24
Remove most global aliases (fix #632)

### LINKED ISSUE #632: Get rid of most or all of the global aliases
Need to write this out to figure out how to write a test for it.  You don't have to read it.  :-D

---

_Project-level_ aliases are fantastic - just so useful in keeping projects readable.  Proven value!  Not on the cutting block.

---

However, we have three other features that overlap now, and I think we can lose one - that one being [global aliases](https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/project/alias_lists.py#L40) - and still be _perfectly backwards compatible_ to this instant.

Global aliases need additional documentation to the user, are inelegant, arbitrary and have been subject to bugs in the past, so I'd love to just lose them.

---

Let's look at a specific example: the alias [`sequence`](https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/project/alias_lists.py#L40).

How can this possibly work??  How can I keep some project that uses `"sequence"` today still working in the future without this line?

Two other features work nicely together.

`python_path` is one - "where do we look to resolve types?".  (Actually, perhaps this is a bad name, because it's not a file path, but something like `bibliopixel.animation` or `bibliopixel.drivers`.)

Users don't need to know about this (though an ambitious user could override it in a project) but there's a path that's passed to classes at construction saying, "Where do we look for contained types?" - and if it's not overridden, it's `bibliopixel.animation.` for animations, `bibliopixel.drivers` for drivers, etc.

The other is "smart resolution" - we handle "files containing a class of approximately the same name" very well (or "files with a single name of the correct type", etc., we're very friendly in opening Python files). So if there's a file my_anim.py, it'll recognized an animation in it that look like `my_anim`, `myanim`, `MyAnim` (as long as there is only one).  It's not much code and it makes tons of cruft just vanish.

---

So suppose I delete the `"sequence"` alias but use it anyway.  What happens?  

Well, after failing to resolve this in the local directory, it tries to resolve this symbol using the python_path `bibliopixel.animation`.  

It resolves to `bibliopixel.animation.sequence` which is already a module, and therefore a namespace, and then uses smart resolution to come up with `bibliopixel.animation.sequence` - same as is already there!

So my theory is I could delete a lot of these outright, and most of the others I could replace with importing symbols from one namespace into another, worst case in __init__.py.

--

The test! 

* For each existing alias:
  * deduce the section (`bibliopixel.animation` etc)
  * run a call to `construct.to_type_constructor` with the alias 
  * `construct.to_type_constructor` is post-alias but uses the python_path from smart resolution
  * check to see that we're getting the same result 
  * ?
  * Profit!

That wasn't so hard once I thought it out.

rec:
This was not entirely a success because I couldn't remove all of them, but it's a good start.  I want to avoid documenting them. :-D

I first write a test that identifies aliases that are redundant due to smart resolution, but leaves them alone.

Then there's code to make more aliases redundant, often by renaming the file to be what I'm now calling "standard form" - where the filename and the classname it exports are "the same" (ignoring case and underscores and the trailing .py) so we get e.g `.devices.dummy.Dummy` and `.animations.off.Off`.

I left variables there to make it all backwards compatible of course.

I found some interesting secrets on the way, such as the fact that the symbol `bibliopixel.layout.matrix.matrix` is in fact... `bibliopixel.layout.matrix_drawing`!  (That fixes this, which isn't 100% backwards compatible, but if someone were using it that way, they were in a state of error...)


adammhaile:
lgtm

PR #635, opened 2018-02-25
Pre-antepenultimate controllers review.

rec:
With this review I'm down to literally one page of uncommitted code that needs to be fleshed out for `animation.controller` - the "last step"

However, I fear that the changes here are pretty abstract to you as so far we still haven't seen the target code, which is still in flux... 

Let me explain what's going on, just to make sure it's on the record.

---

Basically, I ripped apart four shoddy classes and made three better ones of them.  Some renaming occurred.

The following explanation is basically true of "before this change" as well as "after this change" but "after" does more of it with fewer weird classes.

---

Final names are still up for grabs.  Particularly the fact that the base type is `Controls` so I talk about a `Controls` is irksome.  (But there are many control items in it... needs to be resolved...)

---

We're continuing to use the same basic hierarchy - there's an abstract fundamental class `Controls` that handles incoming data of any type, and emit data of any type...

...but all the interesting `Controls` we've written so far are `ExtractedControls` which take an incoming message dictionary `msg` from the source, a dictionary which may have all sorts of crap in it we don't care about, and _extract_ an `OrderedDict` `omsg` with only certain fields in a certain order, normalized.

So the workflow is

* `msg` - "some sort of dict" from the original source (MIDI, keyboard, internet, etc)
  * for example, for MIDI it is mainly just [`vars(msg)`](https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/controllers/midi.py#L56)

* -> "extractor" which emits the ordered `omsg`
  * this is where you throw away the extra crap and normalize the rest.
  * An extractor is defined entirely [by data](https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/controllers/extractor.py#L11-L12).  
  * You can't override any methods on the extractor (this is a positive).  It's very flexible - if you want something else, inherit from `Controls`.

* -> routing
  * This routes the message to set a specific attribute on the target object
  * This is where we use the fact that the message dictionary is ordered
    * We pop off successive elements from the message "from the left side"
    * e.g. in `dict(port='xxx', channel=1, type=control_change, control=2, value=127)` it might be matched by `xxx.1.control_change.2`
    * you can use `omit` to drop some parts - by default `omit` is `port` and `channel`, so the last one would be matched then by `control_change.2`

blah blah blah

BTW, we don't really have to use `OrderedDict`, which isn't totally cheap.  We could be passing around tuples or lists instead with almost exactly the same logic - we'd just have to keep the naming information elsewhere and we only use that in one place - but as far as the base logic goes, the `omsg` is just a list.

I plan to do that later, but `OrderedDict` is really convenient for right now.

On the other hand, this sort of information is likely coming in at a comparatively slow rate - say 100Hz at the most.  The cost of building 100 `OrderedDict`s a second isn't worth losing sleep over.


rec:
Ah, and one more comment.  :-D

Basically, the way I wrote this was to make it really slick for MIDI (which is pretty generic as interfaces go), and then saw what it did to `keyboard`.

Well, I think it ended up being really clear - both [the control adaptor](https://github.com/rec/BiblioPixel/blob/control-2/bibliopixel/controllers/keyboard.py#L22-L51) and the [demo client code](https://github.com/rec/BiblioPixel/blob/control-2/bibliopixel/controllers/keyboard.py#L54-L67) are about as small and clear as you could really imagine.

And actually - I will be able to make it a littel smaller.  :-o 

I really don't need the extractor at all, so in the end stage I'll just make this a `Source`!

PR #636, opened 2018-02-27
Antepenultimate or Controller-related commits

rec:
These deal with the issues that I brought up earlier.

---

The first two commits are tweaks and refactorings.

---

The third one is quite significant - `collection.Collection.index` is now a property, but actually, the whole "Collection that has a `current_animation`" idea got spun off into a separate class, `collection.Indexed`.

`Collection` got too big with this new change, and I think you'll agree it makes no sense for `Parallel` and `Split` - which use multiple animations at once! - to have any idea of `current_animation`.

I didn't just do this to be annoying :-D - I need to be able to say `.animation.index = 0` in the control, and have all the loading of the new image, debug logging of the name, etc occur automatically.

---

And finally, we add a thread-safe event queue to the project.  It's unfortunately only thread-safe with respect to the animation engine, not to the driver engine, so it does not solve the "brightness" issue as I originally advertised!  Oh, well.

It came out nicely.  If you wanted to go right to the event queue suddenly in client code you'd write something like:

    project.deferred_set(Address('.animation.index'), 0)

which would set `project.animation.index` to `0` at the start of the next rendering frame!


rec:
Ah, hold off on this for a bit.  I have a couple of last tweaks, and I just pushed some partial commits to the end of this by mistake...!

rec:
All right!  This actually does something!!!

I rearranged things considerably.  :-D 

Now the first four commits are basically trivial.

Then the split-up of collection.Collection - fun fun.

A series of commits to the Project - adding an event queue, fixing a parser bug, and moving the start and cleanup logic out of main/run.py and into project/project.py.

Then there's a big deal - a final change to Address.  I realized that a lot of the time, when I trigger on some event, at the end of routing I am left with `()`!  So I need a way to add preset values for that very common case - and I did it with the `=` sign so you say things like `animation.index = 1`, which seems pretty clear to me.

The last commit is not so coherent - it's a bunch of small fixes and tweaks in order to actually get the controls seamlessly with projects, and actually, I'm going to tweak that just a hair.  However, there's nothing really alien in there, just a bunch of small  tweaks...

rec:
Ping!

rec:
Me too!  I hope to get out a video with using a fader box and the 12k this week ..... :-o

Issue #645, opened 2018-03-08
Use "standard form" for files and classnames wherever possible.


rec:
"Standard form" means that _if_ a file exists primarily to provide just one class or function, then the filename and the name of the class or function should be "basically the same" (modulo underscores and case).

We want to encourage people to use standard form because it's easier.

A majority of our code is already like this - there are some notable counterexamples like `.animation.BaseAnimation` and `.animation.SimPixel.driver.SimPixel` .

PR #650, opened 2018-03-08
Rename BaseAnimation to Animation (#645)

rec:
I was going to charge ahead and finish #645 but I wanted to run it past you more first.

In particular, names like `BaseCircleAnim` would just go away, to be replaced by `animation.circle.Circle`. 

I think it's really neat and sparse - and I'd leave links back to the old names of course - but wanted to run the whole idea past you.


rec:
I pushed an example of one such commit to the end to see what you think...

rec:
The input needed here isn't really the pull request but on the overall agenda of canonicalizing names (while being backwards compatible).

PR #654, opened 2018-03-09
Give Address an empty constructor and state

rec:
Spun off from the tatters of the general controllers.  :-D

I needed it for a while but I don't really need this at this time, but it's quite elegant to have an empty Address which does nothing and no doubt will be convenient in the future.


rec:
Ping!  This isn't so important but want to get it out of client.

PR #657, opened 2018-03-11
Better deprecation

rec:
I hope this is the last deprecation commit for a while.  It didn't take that long, but it's "technical".

---

I looked at BPA code and realized that deprecation was unsatisfactory, because a majority of the things that were deprecated were still done _just with a comment._

I wanted a mechanism so we can say this:

"If you run your bp 3.x project [with these settings] then it will detect any deprecated features you use"

I found some animations in BPA that still use deprecated imports (our main issue), and set out to try to detect those neatly.

---

Long story short, we can now deprecate things very early - at Python code import time.  We use an environment variable to make the switch (so it goes off "as early as possible").  And if you run

    BIBLIOPIXEL_DEPRECATE=fail bp your-project.json ...

and it doesn't fail, then you are (pretty well) guaranteed that your project will work without changes in BP 4.0.  (There is also the warn setting, but it cannot catch using deprecated imports...)

This deprecation costs something at Python code import time, nothing at frame generation time, and a little at construction time of animations, layouts and that sort of thing (but we do those _once_).


adammhaile:
Note... inline env vars is not possible in Windows. I would prefer `--deprecate fail` or something.

rec:
Ah, I vaguely remembered that nasty catch.  

Well.  Whatever we do has to happen before _any_ bibliopixel symbols are loaded, because we use deprecated in bibliopixel/__init__.py.

So... it has to go into `bp` itself.

Awright then, that isn't really awful, brb.

rec:
Ah, that was more work than I thought but there it is...

PR #666, opened 2018-03-13
Fix serial bug

rec:
The time-consuming part was writing the unit test but finally I understand unittest.mock anyway.


adammhaile:
@rec 
$ flake8
./test/bibliopixel/drivers/serial/driver_test.py:46:13: E123 closing bracket does not match indentation of opening bracket's line

rec:
Fixed!

adammhaile:
ack, will pull and test later tonight.

rec:
So I think you established last night that this is an improvement, so we should merge this and I can continue?

adammhaile:
Yeah, needs more testing, but looks good enough to move forward.

PR #679, opened 2018-03-20
--v4 flag

rec:
The slogan:

"Run your BP3 project with --v4 and if it works, it will work with BP4."


rec:
Yes.  My plan is to remove everything except deprecated.py (which we might use in the future).

That fscking UDP test is timing out again.  I have to figure out a better way...

rec:
Ping!!!

adammhaile:
Sorry, yeah... this is fine. Merge away.

PR #682, opened 2018-03-21
Wait indefinitely for the first result in udp_test.py

rec:
The issue is that the other thread never even starts running before we timeout.

I've changed it so that we wait for the first packet indefinitely and then use the timeout for the rest of them.  I can't prove it, but I'll bet that helps.


rec:
Nope - we got a failure again, though this time we got _most_ of the packets.  :-/

New idea coming.

PR #698, opened 2018-03-26
Disable two flaky Travis tests

rec:
It isn't artnet alone - it's artnet_test and udp_test that hang "alternately".  

My rough estimate is that either of them hangs a little less than one run in 10 on Travis, so Travis fails a little less often than one in five runs.  Of course, we run three times for each commit so...

I actually don't really understand how the observed behavior is even possible.  :-/  So I have little confidence that this will actually solve the problem, but let's see what it does.

By the way, I ran artnet test 50 times in a row here without issue, and I'm now running it 250 times in a row to see if it will hang.  I suspect it will not, which means it probably only happens on Travis, where it's harder to debug.


adammhaile:
+1

rec:
didn't work, but I'm trying more stuff.

adammhaile:
Haha... fair. I kinda wonder how hard it would be to setup another server with multiple python versions, or hell... just 1, that would watch github and run the test suite and then report. Possibly even back to github on the PR. I'm not saying replace travis, just have another.
Actually, once I get the test jig going... I swear, soon (tm), I've been busy... I might set that up to do just that.

adammhaile:
Yep... merge away.

PR #701, opened 2018-03-26
Make layout.color_list setter more flexible

rec:
Ah, I feel bad writing new code when the documentation waits, _but_ this is all part of getting Adam2's thesis project done - which will allow us to accept ArtNet messages as a general control, so that will be super-deluxe!

This makes the setter for Layout.color_list more [procrustean](https://en.wikipedia.org/wiki/Procrustes):  you can either use a list of 3-tuples, or a 1-dimensional list that gets resized.  It also handles lists that are too long or short, and you can call it as a method with an optional offset.

Why is this useful?  Well, the ArtNet data comes in as RGBRGBRGB bytes and this means that I can literally have this routing address: `layout.color_list` _and nothing else_ and it will all work.


adammhaile:
New features is a always why we don't have docs :P But this is for a good reason :)
Looks great. +1

rec:
And this sorta needs to be documented "less" because it will "just work" more often.

It's so much easier if you see "it's cut off" or "it's not long enough" and it works and doesn't fail.

PR #709, opened 2018-03-28
Add numpy to requirements.txt

rec:
I'm not proposing at all to _use_ numpy in the immediate future.

It is however a large dependency and I'd like to find out if this is an issue for anyone in installation before we grow to become reliant upon it. 

So I'd like numpy to be a dependency for at least a few cycles before we actually require it for anything... so now's the chance.


adammhaile:
+1

PR #712, opened 2018-03-29
Make Layouts cloneable and other final tweaks to controllers.

rec:
I am pleased to announced that I have reached my holy grail, fading backwards and forwards between animations with a MIDI controller, and this is the pull request with the last details.

I think it's particularly slick.  On the Mac at least, you can plug in your controller that sends continuous controllers 0, 1, 2, 3 (which is basically all of them out of the box) on any channel, MIDI device etc

    pip install mido  # one time, or we could add it as a dependency
    bp projects/23-midi-control.json

and you're off - it autodetects you have just this MIDI device, etc so no configuration necessary.

I'm going to have to make a movie of this.  :-D


adammhaile:
DO IT!

PR #713, opened 2018-03-29
Allow animations to be retrieved by name.

rec:
Two main parts;

* I make sure that each animation at each level has a unique name
* then I allow you to retrieve the animation by name or by index

Some bonus magic lets you retrieve it also as an attribute so you can say `animations.Wave`, so convenient, see the examples.


adammhaile:
+1

PR #714, opened 2018-03-30
Allow controls to raise exceptions (fix #711)

### LINKED ISSUE #711: Give controls the option to throw an exception and die
Right now, I basically swallow all exceptions in a Control and just report the exception a few times, and then nothing.

For production, this is great.

For debugging, this is very inconvenient.

rec:
I'm actually debating changing the default behavior to be `'raise'` because it's amazing how quick the eye is to ignore errors at times...


adammhaile:
+1

PR #730, opened 2018-04-04
Command line flags override project files (fix #694)

### LINKED ISSUE #694: Command line flags should override project files, not the other way around.
What's the point of a command line flag if it can't override the hardcoded value?

It's basically switching the first two parameters on this line: https://github.com/ManiacalLabs/BiblioPixel/blob/dev/bibliopixel/main/common_flags.py#L170
but there would no doubt be other tweaks needed to defaults and even some logic.

rec:
* Surprised it's as easy as that

* I suspect there might be some problems with all these command line options and
  their interactions that I have not found, but we'll test all this before the next release

* Automated testing is hard for this


adammhaile:
+1

PR #732, opened 2018-04-04
Fix unit test breakage due to loady change

rec:
Ah, I made loady a little cleverer and broke an edge case in a bp test - here's the fix.


adammhaile:
+1

rec:
Ah, this just isn't good enough.  I need to tweak loady or every branch will fail.  Sigh.  brb

adammhaile:
Setup travis for loady such that it does that?

PR #736, opened 2018-04-05
v3.4.12

rec:
Note that I didn't actually announce v3.4.11 because I immediately discovered the print statement.

This is a downside from the fact that my standard test spews all those GIF errors to the terminal, where they appear in red, catching my fallible eye...

Also, we really have no good way to test for this issue (useless print statements).  Further vigilance is needed.


adammhaile:
+1
Sorry, I'll try to fix the gif errors tonight.

rec:
Ah, it's more my fault than yours.  :-)

As you know, I always look for some way to avoid errors in future.  So how to prevent "leaving a print statement in the code"?  This is the third time this has happened to me on this project in the last year!

Answer - really none, unless we also have integration tests that test all the output from a run.  We don't have the manpower to really get into that.  Ah well!

adammhaile:
`pip install flake8-print`

Then flake8 will emit T001 error code if it finds any print statements.
I'm 100% ok with just not allowing print()

PR #737, opened 2018-04-05
Remove or mark with noqa all instances of `print`

rec:
* Eliminate all style violations found by `flake8-print`
* Now we can use `print` for debugging and have it caught by our flake8


rec:
It also means that we can later easily redirect all output by redirecting `log.printer`.

rec:
Good idea!

rec:
Let me see if that works with Travis, and then I'll push it.

adammhaile:
Cool

PR #743, opened 2018-04-07
More signal handling

rec:
I still haven't figured out why we don't see changes in project files on a HUP!  It's very weird, we do seem to reread from the filesystem...!

But I checked in a script that lets me reproduce Adam2's issue, which is halfway to solving it.

Also - new features!

For this purpose, now when `bp` is running a Project, it stores its PID in a well-known tempfile.

So you can now say things like:  

* `bp pid` - print the PID for the current project
* `bp signal SIGHUP` - restart the current project
* `bp signal` - same (SIGHUP is the default)
* `bp signal SIGINT` - shut down the current project

or any signal you like - though we only do something intelligent for `SIGHUP`, `SIGINT` and `SIGTERM` as of today.


adammhaile:
+1

PR #745, opened 2018-04-08
Fix `bp run --dump` option

rec:
Oops.  `--dump` was broken because there are now objects in the tree being dumped.  Now we use `repr` to represent such things.


adammhaile:
+1 nice catch

PR #746, opened 2018-04-08
v3.4.13

rec:
In particular, Adam2's issue with not rereading project files was traced down to a caching issue in loady (/me rolls eyes), so a new version will trigger an update of loady to the new version.


adammhaile:
+1

PR #749, opened 2018-04-08
Add RemoteServer.shutdown_server() classmethod and SIGHUP fix

adammhaile:
@rec - Ok, this should take care of you.
From anywhere just call RemoteServer.shutdown_server() (it's a classmethod)
This uses the "official" method for shutting down. I say official in quotes as their comment on the subject does not instill confidence: https://github.com/pallets/werkzeug/blob/b96829e551cb8ad88c91a89e2e1fa154368065ad/werkzeug/serving.py#L304
Everything will shutdown and then you'll have to call run_server() again... or if you need make run_server() auto restart if that suits your needs.

Also, this should take care of SIGHUP on Windows unless I'm understanding wrong. I mean, it works on Windows at least for trying remote_demo.json
Basically Windows doesn't have SIGHUP so this update auto ignores processing of any signal not available on the system.


rec:
Are you sure this will really work??  Do you know for sure that `signal.SIGHUP` is not defined on Windows?

I think I'd rather just have two different lists of signals depending on `platform`... then we would know exactly what was going on...

bibliopixel/animation/remote/server.py
```diff
@@ -23,6 +23,14 @@ def __init__(self, q_send, q_recv):
         self.q_send = q_send
         self.q_recv = q_recv
 
+    @classmethod
+    def shutdown_server(cls):
+        log.info('Shutting down UI server.')
+        func = flask.request.environ.get('werkzeug.server.shutdown')
+        if func is None:
```
rec: This should just be a warning - or even it could just return False if it didn't manage to shut down.

bibliopixel/animation/remote/server.py
```diff
@@ -23,6 +23,15 @@ def __init__(self, q_send, q_recv):
         self.q_send = q_send
         self.q_recv = q_recv
 
+    @classmethod
+    def shutdown_server(cls):
+        log.warning('Shutting down UI server.')
+        func = flask.request.environ.get('werkzeug.server.shutdown')
+        if func is None:
```
rec: Tiny quibble if you cared:  you could rewrite this as

    return func and (func() or True)
     
or clearer but longer:

    if func:
        func()
        return True

...or... not.  :-D

adammhaile:
Mine's clearer :P Keep your python black magic to yourself.

PR #757, opened 2018-04-09
layout.Matrix now stores parameters needed for cloning

rec:
Things are coming into a nice rhythm.

Basically, a nicely behaved object created in a project will have attributes that are named the same as its construction parameters!

In this case, Matrix had construction parameters `vert_flip`, `y_flip` and `serpentine` which weren't stored, so it couldn't be cloned (which is what we need for any of the parallel animations).

This fixes that.


adammhaile:
+1

PR #759, opened 2018-04-09
New Limiter animation (fix #708)

### LINKED ISSUE #708: Device brightness limiters/compressors
From [the list](https://groups.google.com/forum/#!topic/maniacal-labs-users/0Qe75tYnBc8):

> I'm also looking to limit the total power consumption of each pendant so that I don't have to have each light hanging by a thick cord. By treating each pendant as a unique object, I hope to be able to set a max total brightness, which would be used to to automatically scale down the array for each pendant as needed without compromising the brightness of a single lit pixel.

rec:
The first commit is basically a bugfix.

Then we simplify the math handling considerably, removing a whole poorly-conceived `ColorList` class and replacing it with nothing.  Much better!

The rest is self-explanatory.


adammhaile:
Can you explain what Limiter is for?

rec:
Ping!

I'd like to push a new version today and get this in.

adammhaile:
Do it

PR #761, opened 2018-04-09
New `bp restart` and `bp shutdown` commands replace `bp signal`

rec:
Again, from the docs branch.

I realized that explaining `bp signal` was dumb, and instead I have commands `bp restart` and `bp shutdown`.  You can still send another signal from those if you want... :-)


adammhaile:
+1

PR #764, opened 2018-04-11
v3.4.15

rec:
I want this new version so the `bp restart` and `bp kill` commands replace the old `bp signal` so no one ever needs to know about the old command.  :-)


adammhaile:
+1

PR #769, opened 2018-04-18
New `Runnable` interface + grab bag of tiny commits

rec:
My big artnet branch got out of control and I had to extract out a lot of simple commits.

The "Runnable" one is the one that needs explanation:

As I get more UDP tests, I'm having significant issues with sockets not shutting down correctly and this problem is only going to get worse as we have more facilities.  The main artnet branch is stymied now because of this.

In the real world, these will correspond to `bp` refusing to shut down and having to be killed.

So I extracted out a class called `Runnable` that basically abstracts a `Thread` and gives it a bit more structure, as well as forcing all joins to have a timeout.  

It's useful both for threads, and for other services that have their own threads hidden within them (like `pynput` and `remote`).


adammhaile:
+1!

PR #782, opened 2018-05-03
Thread vevamp; ArtNet control

rec:
Because this branch got too large, I split off a bunch of uncontroversial bug fix or cleanup commits out at the start.

The last two are more interesting.

The second-to-last one, "Rework Runnable and threads for better shutdown" is about as ugly as I have written on this project.  :-/  

I needed to change the threading model somewhat to deal with some of the issues I recently documented and there was no neat way to do it gradually.

I did test everything really thoroughly.

It's all worth it,. though, for the last one: "New ArtNet control".  The code for the ArtNet control is [tiny and very clear](https://github.com/rec/BiblioPixel/blob/artnet/bibliopixel/control/artnet.py).

---

I haven't yet tested the ArtNet control - that last commit - on Adam2's ArtNet installation, although I have really a nice integration test that took a lot of work to really make sing.  

I'm going to test the hardware in parallel with the review of this branch, or if the branch gets finished before then, I'll pull out that last commit into a separate branch and pull all the other parts.


adammhaile:
+1

PR #789, opened 2018-05-04
Remove test that crashes ArtNet receiver

rec:
GAH.

I did not intend to check in ArtNet until I'd run it under fire, but I got distracted.  :-/


adammhaile:
+1

PR #816, opened 2018-05-21
Now unknown animation types are replaced with Failure (fix #663)

### LINKED ISSUE #663: Failure to load an animation with Remote is a hard fail
It used to be that if running remote and an animation failed to load, it would just skip it and grey out the animation button. Now it just hard fails. Tested when trying to run this project file: https://github.com/adammhaile/BiblioPixelProjects/blob/master/SparkCon2017/1K.json
Failed on ImageAnim because pillow was not installed.

rec:
We were previously catching the case with a known `"typename"` which failed to construct.  We were not catching the case where the `"typename"` is missing or can't be imported.


adammhaile:
Checks Failed :( Does this still work with Remote?

rec:
Ooops, pushed a fixup.

Absolutely, I in fact tested both types of failure with remote:

* Bad typename
* existing typename with bad constructor parameters.

adammhaile:
lgtm

PR #818, opened 2018-05-21
Better reporting of bad ports for SimPixel connection (fix #720)

### LINKED ISSUE #720: Better reporting of bad ports for SimPixel connection.
> Currently I am working on my local machine (MacBook). I have been able to change the port to other four-digit port numbers and display on SimPixel just fine. However, when I choose port 80 for the driver in my project config, I get this cryptic error message:

`ERROR: 'NoneType' object has no attribute 'update'`

> Why might port 80 not be valid? 

(Answer: "you really can't use port 80. Unix-based systems disallow binding to any port below 1024 for non-root users.")

We need to report such errors better!

rec:
Now there's a much better error (if not in -v mode):

    SimPixel attempted to open a port at 80 and got a PermissionError.

    Many systems disallow binding to any port below 1024 for non-root users.
    Try again with a port number that's 1024 or greater.

I felt this was a decent blend between "too much information" and not enough.


bibliopixel/drivers/SimPixel/websocket.py
```diff
@@ -81,3 +87,11 @@ def remove_client(self, client):
             self.clients.remove(client)
         except:
             pass
+
+
+PORT_BELOW_1024_ERROR = """
```
adammhaile: Move above it's usage. I know python lets you do this. But I look *up* for variables, not down.

rec: I'll change it, but there's a reason for this.

Basically, I go from "interesting to boring" - because I am writing for the reader of the code - who will be "me" in a few weeks as I forget this.

I want to start off with the key, must-know, seriously-important stuff at the very top, and then put boring things at the end.

The reason that `PORT_BELOW_1024_ERROR` is at the very bottom is that it's about the most boring thing there.  When you see `PORT_BELOW_1024_ERROR` in the code above, you don't say, "Oh, my gosh, what could that be? I need to know!"  :-D  

So to me, putting it at the top of the page forces the reader to read this boring error message _first_ before the actual important stuff happens.

PR #821, opened 2018-05-21
Allow serial driver to hardcode self.dev (fix #802)

### LINKED ISSUE #802: Serial(dev='/dev/ttyACM0') no longer works
BP2 had this: https://github.com/ManiacalLabs/BiblioPixel2/blob/master/bibliopixel/drivers/serial_driver.py#L170
BP3 has no equivalent :(

rec:
Needs more testing...


rec:
Pushed a fix to your first comment!

PR #825, opened 2018-05-23
Make data_maker more convenient for users

rec:
* User-friendly type alias data_maker.NumpyColorList
* Simplify imports of numpy (since it's now a dependency)
* This came from the documentation branch


adammhaile:
+1

PR #837, opened 2018-05-28
Add $ to existing @ as introducers for aliases (fix #836)

### LINKED ISSUE #836: Use a symbol for aliases that Yaml understands to be part of a string.
Right now we have to quote all our strings with aliases in them, because `@` is some special character for Yaml.

If there is no such character, then document this issue.

rec:
I'm going to use `$` in all the examples in the documentation as then I don't have to quote the strings...


adammhaile:
+1

PR #846, opened 2018-06-08
Fix issue that prevented projects from running

rec:
* Introduced in 4e39191
* exposed in 9441a67
* We could rebase this into 4e39191 but I am weakly against doing this


rec:
It failed only on Python 3.5, none of the others.  However, the traceback seemed like it was actually missing a datafile in a unit test.  When it reran, it was fine.

I chalked it up to weirdness.  The datafile is definitely there, and why would it not be there just for 3.5?

rec:
Ah, is there a way to see the log for previously failed tests?  I unfortunately closed that window... :-/

adammhaile:
Why would calling a different load function change that?

rec:
Oh, er, sorry! Wrong answer to the right question.

We had an intermittent Travis failure in this pull request which I had just cleared up - so I thought you were talking about that.  Context error.

The problem this fixes is simple - as of this instant, you can't run any projects at all in dev because you get a traceback.

adammhaile:
Oh, cool. +1 Carry on.
Also, old Travis CI logs are here: https://travis-ci.org/ManiacalLabs/BiblioPixel/builds

PR #847, opened 2018-06-09
Rename project/event_queue to project/edit_queue

rec:
Another "change the code to match the documentation" commit.  

It isn't a queue of "events" at all - one incoming event can trigger multiple of these!  These are changes to the project, so they are an "edit" - pending a better name.  :-D

(Also, we actually use `threading.Event`, which means we have two unrelated `Event`s.)


adammhaile:
+1

PR #858, opened 2018-06-27
Cosmo docs

adammhaile:
Pulling in @MyNameIsCosmo tweaks and MkDocs setup. @rec - This one's for you :) I have more work to go here to get it going the way I want but that will come quickly.


adammhaile:
Self approve!

PR #859, opened 2018-06-29
Use safe loading and dumping with yaml

rec:


rec:
I actually _rebased_ the docs branch against dev, so be aware that I rewrote history!

adammhaile:
Ah... good to know. Thanks

PR #865, opened 2018-07-06
Rest Control now responds to PUT method (fix #863)

### LINKED ISSUE #863: New rest client uses get requests for everything?
Just noting that it could just use an actual GET or PUT request at

localhost:8787/address that either gets or sets the variable

as of right now it uses a GET command to an address that says /put/address (which is find, and it's not actually interacting with data so doesn't need to be restful, but should document that all requests are GET requests then.

rec:
I made the fix, which was tiny, but I also wrote a little integration test for the rest server, so anyone can make fixes to it and run the test.


estiens:
Just as a total nitpick but a truly RESTful service has different actions at the same endpoint -- so for example in this case if you went to localhost:8787/animation.brightness with a GET command, it would return the current value, if you went to the same endpoint with a PATCH/PUT command, it should update that value, etc

I do like a fallback option of only being able to make GET requests to trigger a value change, but that's not actually RESTful at all, and it should be on another endpoint

so something like 

GET localhost:8787/animation.index (returns 1)
PUT localhost:8787/animation.index with {value: 2} as the params (changes animation index to 2)
(curl -X PUT -i 'http://localhost:8787/animation.index' --data value=2)

fallback GET localhost:8787/update_value/animation.index/2

--

Also worth noting that as written, you have to use a different endpoint for each value which means this control can only change one thing as a time. You could also envision something like

PUT  localhost:8787/animation, params: {brightness: 20, color: 120, fps: 5} as being an update endpoint that can update multiple values of a running animation simultaneously.

Just suggestions, nothing wrong with the way it is written, if I have a need for these things in current project will try to push up a small PR
G

adammhaile:
I have no issue with the "correct" RESTful way being an option but the
GET-only way is a requirement. Truly RESTful APIs are for developers who
understand such things, which our users are not.

estiens:
Fair enough! Thanks for implementing this! Mostly just needed the ability to set multiple values at a time with one HTTP call. Works a charm :)

PR #866, opened 2018-07-07
Make the Rest control actually REST compliant (fix #863)

### LINKED ISSUE #863: New rest client uses get requests for everything?

rec:
Worked right the first time!

I also spent a great deal of time documenting this, so the documentation for this control should be complete.


rec:
@estiens

PR #867, opened 2018-07-08
For discussion: Refactor serial/devices.py and driver.py

rec:
The new code is somewhat clearer. 

More, it points out that under some codepaths, we can crash (raise an exception that is never handled) after initialization if any driver ever reports an error.

It is my belief that we shouldn't ever crash.  Perhaps it might be OK on construction of a driver (even then I don't think so), but certainly not just because we get an error during run.  It might be that there are several drivers and one gets unplugged, or simply a little line noise for a few minutes - it shouldn't bring the whole thing to a sticky end...


adammhaile:
I'm ok with all of this as is, barring that it gets extensive testing before any release of course.

rec:
Why don't we leave this until we get the test harnesses, then?

BTW, do you have a good name for those?

adammhaile:
Fair... finished assembling yours last night. Flashed the Pi with OS. Will
test again to make sure it all works, then send out.
No... no real name. It's the Maniacal Labs BiblioPixel / AllPixel / etc LED
test rig I guess :P

PR #871, opened 2018-07-13
Convenient animated GIFs

rec:
We need this to do the documentation properly.  

I want to have lots of examples of projects with embedded animated GIFs to show what happens, and now I can make them wholesale, and remake them 

Now you can just say `bp --gif your-project.yml` and it will write out an animated GIF called `your-project.gif` that's ten seconds long.  Or you can use JSON for finer control:

    bp --gif="{filename:  stuff.gif, time:30, render: {pixel_width: 8, vertical: true}}"  your-project.yml


adammhaile:
Haha... this has been written in various ways various times. But I like
this. Good stuff.

PR #877, opened 2018-07-16
 New `bp new` command makes a new project directory from templates

rec:
@estiens


bibliopixel/main/new.py
```diff
@@ -0,0 +1,109 @@
...
+    # Create a new project directory named my-project/ in the current directory
+    bp new my-project
+
+    # Create a new project directory named my-project/ in ~/projects
+    bp new my-project ~/projects
+```
+"""
+
+LETTERS = 'abcdefghijklmnopqrstuvwxyz'
+LETTERS += LETTERS.upper()
```
adammhaile: import string
string.ascii_uppercase

rec:
This will be, like, the first page of the documentation.

:-D

bibliopixel/main/template/animation.py.tmpl
```diff
@@ -0,0 +1,45 @@
...
+            color=COLORS.green,
+
+            # End of the fields for your class
+            **kwds):
+        super().__init__(*args, **kwds)
+        self.color = color
+        # Your initialization code goes here.
+
+    def step(self, amt=1):
+        this_pixel = self.cur_step % len(self.color_list)
```
estiens: but wait, i'm already confused as a newbie. We jut set a variable called `self.color` and now we are finding the length of `self.color_list` Clearly one of these variables (at least) comes from a parent class. Do we always have access to self.color_list? If we set `self.color` does it automatically go to the `color_list` etc.

PR #886, opened 2018-07-18
New colors.to_color, and unittests

rec:
Spring cleaning!

Extracted this from an abandoned branch.

`to_color` gets used presently... but also, all those tests...


rec:
OK, I just added a more controversial change to this!

I need to fix `Fill`, which is becoming pretty key, and to do so I need features that are only in the dev branch of BiblioPixel.

I'm not going to move every random Animation in, but this one is so key to almost every animation with over a few dozen components...

rec:
Ping!

I swear, I won't move anything else from BPA here - just this one, because it is so useful.

PR #889, opened 2018-07-23
Release test

rec:
The long awaited test of the test board!!  :-D

The first four commits are basically trivial and independent, the last one is pretty huge.

The program discovers the features that your specific platform supports, reports on the ones it cannot, and then runs the rest.  (There's a `--features` flag to override this.)

You can specify which tests to run on the command line, or if there are none, it runs all the tests.

The test case for the board is called "all_pixel" because that's what gets detected.

---

If I get the spare time, I'll bring this within the unit test framework using a feature flag there (we already have these), but it was faster to do it this way for the moment.


adammhaile:
Please provide example invocation... you know, docs ;)

rec:
Check it:  https://github.com/rec/BiblioPixel/tree/release_test_2/scripts/release_test

rec:
Approve this and then I can test the final 3.4.19 with the board and all, and then wait for your approval!

PR #896, opened 2018-07-26
Tweaks to help

rec:


rec:
(Came from writing dox...)

PR #897, opened 2018-07-27
Fix fill.Fill

rec:
This never worked as checked in.

I brought this from a bigger branch and clearly didn't test it at all.  :-/ :-/ 

Sorry, too many things.  I'll do better next time.


adammhaile:
+1

PR #901, opened 2018-08-03
Support Python 3.7

rec:
All the dependencies are now ready - with the single exception of Travis.

Given that people are automatically downloading 3.7 on Mac and Windows, I'd like to push this before Travis is ready, and then I'll do the final tweak when they are.


adammhaile:
Travis doesn't support 3.7 yet? Lame.
+1

PR #909, opened 2018-08-08
Delete commands from `bp` with feature flag `BP_DELETED_COMMANDS`

rec:


adammhaile:
I'm confused as to what's going on here.

rec:
There are two groups of commands that we should probably delete.

`bp update` doesn't work as well as `pip3 install -U bibliopixel` does.

The user default commands change these hidden preferences files in your home directory and are dangerous, and offer nothing over `bp defaults.yml+project.yml`.

But I might be wrong.  So I didn't delete those commands - I just disabled them with a feature flag, one that we could turn on again if we liked.

rec:
Ping!

adammhaile:
+1

rec:
My theory is that no one will want these commands back.  If they do, we can tell 'em to set that environment variable temporarily... :-)

PR #913, opened 2018-08-10
Encapsulate time functions and mock them out when rendering

rec:
I gathered together `time.sleep()` and `time.time()` calls and moved them onto `Project`.

Then I added a switch that lets you load a `FakeClock` where `sleep` takes no time at all.

Result - speed up rendering GIFs by about 80%.


adammhaile:
nice :) +1

PR #916, opened 2018-08-13
Gif improvements

rec:
Pulled from the docs branch.

The usual crop of small changes, and then the last one has three solid ones.

* Now GIFs all have approximately the same width (512 pixels or a little more)

* You will be pleased to know that we have so many GIFs being rendered in the docs branch that I had to write code to cache them so we didn't recompute them each time!

* Slight improvements in pattern matching gif file names

* All the computed .gif and .gif.yml are hidden in a subdirectory - but we will be checking them in

The documents are pretty spectacular these days, and I am getting momentum...


adammhaile:
+1

rec:
BTW, Github complained that the images are too large (>50MB), so I tweaked the image size to be 384 pixels or a little more.  We can always adjust that at the end!

adammhaile:
Oh... shit. What? Why are they so large?
We should never include files that big... we will forever be stuck with
them in the history. They should be hosted elsewhere.

PR #922, opened 2018-08-15
New run.repeats simplifies run Section

rec:
`run.repeats` is completely backwards compatible, but is also easy to explain


bibliopixel/animation/runner.py
```diff
@@ -62,6 +66,10 @@ def __init__(self, *, amt=1, fps=0, sleep_time=0, max_steps=0,
         self.threaded = threaded
         self.flat_out = flat_out
         self.main = load.code(main)
+        if repeats is not None:
```
adammhaile: This will need to be made clear that this ONLY works if the animation supports it. Otherwise this does nothing and the animation runs forever.

rec: I'm not sure what you mean?  This translates exactly into the old values.  Nothing changes...

bibliopixel/animation/runner.py
```diff
@@ -34,6 +35,9 @@ def __init__(self, *, amt=1, fps=0, sleep_time=0, max_steps=0,
         if fps < 0:
             log.error('fps %s < 0', fps)
             fps = 0
+        if repeats and repeats < 0:
+            log.error('repeats %s < 0', fps)
```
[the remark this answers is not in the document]
rec: Ooops, fixed.

adammhaile:
I'm just saying that the doc MUST specify that this ONLY works if the
animation you are calling sets state = STATES.complete at some point and
link out to the docs for animations that show how to do that.

rec:
Ah, that's a really good reminder.  I did "know" that but wouldn't have put it in the docs.

adammhaile:
Oops. Sorry, I forgot that I needed to go back and approve it.

PR #931, opened 2018-08-20
Allow `main.run` filenames to be Yaml (fix #929)

### LINKED ISSUE #929: `-j`/`--json` flag got lost...
It used to allow you to enter project data as a JSON argument from the command line.

There is code there to define it, but no code reads the flag!  I know I had this working at some point in the past but not super motivated to track down when I accidentally killed it.  :-)

An upcoming pull request will repurpose that flag.  I'm leaving this issue, because we should make this part of the "run" loading system so you can say things like:

    bp my-setup.yml + "{layout: {serpentine: false}}" + my-project.yml

rec:
That was short and worked right the first time.  

We can now write things like `bp projects/23-midi-control.yml + '{shape: 100}'` - isn't that nice?


adammhaile:
+1

PR #934, opened 2018-08-21
Recognize class names in command line

rec:
From docs work.

This is very slick - now you can call animations from the command line by their typename!

Just write:

    bp BiblioPixelAnimations.matrix.Twinkle

or

    bp '$bpa.matrix.Twinkle'

or

    bp @bpa.matrix.Twinkle

Works with `+` notation, etc.

So now you do `bp animations` and then pick one to run - done.


adammhaile:
+1

rec:
Ah, breakage, one moment... :-/

PR #935, opened 2018-08-22
Active index page for REST server

rec:
I also cleaned up the REST server considerably so it now handles errors a lot better.

Here's the page after it got or set data successfully:

screen shot 2018-08-22 at 16 45 31

Here's the error page:

screen shot 2018-08-22 at 16 45 56

@estiens


adammhaile:
Nice! +1

rec:
Ah, I just pushed a last change at the end after you approved.  :-D

A new class called `rest.OpenPage` that is the Rest control that opens the main page.

Why?  So you can do this:

    bp my-project.yml + '{controls: rest.OpenPage}'

and it will immediately pop up the rest page to play with...

adammhaile:
+1

PR #936, opened 2018-08-23
New `scripts/bp-pid` prints PID of `bp` process

rec:
* If there is no such process, it returns an error

* Very fast; loads only a handful of Python libraries

* The intention is to deprecate `bp restart` and its buddies in favor of
  `kill -hup `bp-pid``

This means four fewer commands I have to document.  :-D

This also solves the issue that came up on the list of "wanting to restart `bp` without calling `bp`".


adammhaile:
+1 except of course it's not windows compatible :P

rec:
No less incompatible than before!  :-)

adammhaile:
Fair enough :P

PR #937, opened 2018-08-26
Remove unused, non-working file

rec:


adammhaile:
Why did this file break???? We should not be breaking compatibility with
existing projects

PR #943, opened 2018-08-27
Fix most of the release_tests

rec:
Fixes everything except the MIDI test.

MIDI is not broken - however, it uses a Mixer which breaks other animations that are in it, and I want to resolve that.


adammhaile:
+1

PR #945, opened 2018-08-30
Allow any layout for strip animations

rec:
I'm trying to add little matrix animations to the bottom of each documentation page - but then many of my choices don't work just because of this warning.

Is there a reason for this warning?  If we can't delete it, can we at least add a constructor parameter to allow this?

Pretty well every strip animation looks really good on a matrix.  I'd really like to figure out how to allow them to be used...


adammhaile:
The intent is that Strip animations aren't always going to and aren't supposed to look good on a matrix. As the matrix could have any variety of layout. It's just to keep you from doing something silly.
I'm ok with making this a non-failing warning, but don't just remove it. I at least want to yell at the user in case they didn't realize there's a difference.

rec:
All right!  I was actually able to pull that test up to `Animation.animation.__init__`, removing a lot of duplication.

rec:
All right!  I was able to pull that up onto `Animation.__init__`, and so give better warning messages while still decreasing total lines of code!

rec:
Fixed!  It's still less than the original code.  :-)

rec:
Ping!

Shorter, better error messages, fun fun fun.

bibliopixel/animation/strip.py
```diff
@@ -5,14 +5,11 @@
 class BaseStripAnim(Animation):
     LAYOUT_CLASS = Strip
     LAYOUT_ARGS = 'num',
+    FAIL_ON_WRONG_LAYOUT = False
```
adammhaile: Please set this in BaseStripAnim. I don't like magic values that only appear when needed and for which their default is in some random line of code. I understand this is entirely "valid" but it makes me nauseous and it makes the code VERY hard to read and understand.

rec: Done!

adammhaile:
Thanks +1

PR #951, opened 2018-08-31
Adding __spec__ for MP

adammhaile:
@rec - Fix for Windows MP issues.


rec:
:+1:

Whoa!

Why does this work?

PR #953, opened 2018-08-31
Rename log.warn to log.warning in animation.py

rec:
Oops.  :-)


rec:
(I should add that even if someone were using that commit ID, it wouldn't have caused a working project to fail - only a project that already failed to fail in a different way...)

adammhaile:
+1

PR #963, opened 2018-09-04
release test tweak

rec:
The new `bp` release test doesn't work when verbose is off!

Fixed that, tweaked the output slightly.


adammhaile:
+1

PR #964, opened 2018-09-04
A little work on test coverage...

rec:
Super boring. 

I hadn't looked at the coverage in months so I did.  About where we were before, but I managed to get rid of over a third of the file names appearing in the report with a few minutes' work, though it didn't increase our coverage score very much...


adammhaile:
+1

PR #971, opened 2018-09-16
Add mido to test_requirements.txt

rec:
Re-testing everything in a clean virtualenv, and this dependency is missing.  

(You wouldn't have noticed it as it detects you don't have MIDI.)


adammhaile:
+1

PR #972, opened 2018-09-16
Add mido and rtmidi to developer_install

rec:


rec:
This is the correct way to do it!

adammhaile:
+1

PR #974, opened 2018-09-16
Add a Python 3.7 classifier to setup.py

rec:
Ah, does this mean a new minor version?  :-/


adammhaile:
Well... yes. Technically.

PR #979, opened 2018-09-21
Fix deprecation warnings from newest pytest

rec:
The newest version of pytest has more deprecation reporting.

Ours tests inconsistently used `assertEqual` and deprecated `assertEquals` - I fixed this.

The warnings in all_test are from third-party libraries so I silenced them.


rec:
ah, sorry, will fix... I extracted it from another branch.

rec:
Fixed.

PR #981, opened 2018-09-22
New Palette class (fix #978)

### LINKED ISSUE #978: Color palette system for animations
Numerous animations have hard-coded color palettes - too many to list.

Others have some sort of color system but set it haphazardly.

We need a new type called Palette which could eventually be:

* a single color
* a list of discrete colors
* a continuous (linear) gradient of colors
* a surface or volume of colors (2- or 3-d gradients)  (EDIT: that feature was too hard and is deferred)

Actually, we can expand the existing type "Colors" which already does some of this!

rec:
While this fixes the issue as written, the other half will be https://github.com/ManiacalLabs/BiblioPixelAnimations/issues/41 (after this is released).


bibliopixel/util/colors/palettes.py
```diff
@@ -0,0 +1,40 @@
...
+    'pastel': _colors('azure 3, eggshell, sienna 2, goldenrod'),
+    'tints': _colors('plum 4, dark grey, antique white 2, linen'),
+    'splash': _colors('steel blue 3, gainsboro, gold, tan 2'),
+    'energy': _colors('magenta, yellow, cyan, beige'),
+}
+
+DEFAULT = 'rainbow'
+
+
+def get(name='rainbow'):
```
adammhaile: Doc on this method?

scripts/nearest-color
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
adammhaile: What is this used for?

rec: This isn't really very important - I wrote this when I was extracting the palettes and didn't want to entirely throw the code away.

(I didn't want to exactly steal palettes from other sites, so I took palettes that existed and converted them to the nearest named colors.)

I'd be very willing to drop this, but my plan was actually to remove `bp color` (because I'm wanting to do less in `bp`) and merge it into this.

bibliopixel/util/colors/palettes.py
rec: Ah, I think I need a little more documentation in general, and this method in particular.  I'll write it up and then get back to you quite soon...

rec:
All right!  I added a lot more clarifying documentation and changed some naming slightly - you now get a color from a _position_ in a Palette.  (I realized that `index` wasn't quite right, because an index should be an integer, and this is expected to be a float...)

scripts/nearest-color
adammhaile: ok.... I just wanted to make sure it was clear what this script was for. Please at least add a usage note (include why you would use it). But fine with keeping it.

rec:
Done!  Waiting for the CI...

PR #990, opened 2018-10-01
v3.4.23

rec:
I'll push this some time tomorrow after the hotfix goes off...


rec:
Updated.  I'll push this in an hour or two.

PR #994, opened 2018-10-03
Better release tests

rec:
This would have detected the last two issues...


rec:
My plan is to push this tomorrow, and then make a new release with actual stuff in it, not just hotfixes.  :-D

adammhaile:
Sounds good. +1

PR #1005, opened 2018-10-12
Preparation for "envelopes and LFOs" branch

rec:
This is from my other working branch, "envelopes and LFOs".

These commits are also over a month old, so I extracted out the fully tested parts so I have less work hanging out in random branches (which I find stressful)


ozel:
that looks super useful! How would it be used in a project file?

PR #1035, opened 2018-10-30
Preliminary steps to POV fix

rec:
Early steps while I contemplate how to do the frame multiplexing trick.  It's not hard - the question is how to present it to the user?


adammhaile:
I don't understand what `direction` has to do with POV?

rec:
Right now, POV only scans the matrix in one direction - top to bottom, if I remember the code aright.

But there are four possible directions it could actually scan.  

Experimenting with the two `_flip` variables shows that you can't get to two of them that way - specifically, if the strip were hung horizontally so you moved your head up and down, rather than vertically (or do I have that backwards?)

(Also, to be frank, I'm always looking to add variables that are meaningful so I can add a control to them.  :-D  In a square matrix, attaching a fader to `direction` and moving it around would possibily have mind-blowing effects - or perhaps not, but one way to find out.

adammhaile:
ah.... So, a little feature creep.
POV was always sort of written in a way such that there was one correct way
to use it. Not saying it should be expanded, just not sure it's worth doing
it just for the sake of it.

rec:
I agree, to be honest.  :-D

In fact, I was doing this as "doodling" while try to figure out the right way to fix the main POV.

The issue is that it's a layout, and layouts aren't supposed to be doing temporal things!  Layouts are supposed to be passive, "index remappy" sort of things.

I'm leaning towards turning POV into a container Animation at this moment, tbh, even though it isn't backwards compatible.  At least we can get really clean coherent timing that way - but it's not perfect either.

Or - logically, POV should be a driver - a driver containing another driver?  

In other words, the POV matrix driver would contain say an LPD8806 strip driver.

We have no such thing until now but there's no particular reason that such a thing shouldn't exist or difficulty in doing it.

But then we have the issue that we again have two different "frame" generators that won't be in synch with each other.

rec:
The container Animation is, I have decided, the only correct way to go.  Here's why.

I might very easily want to have a strip Project with a bunch of different animations controlled by Remote or REST, only one of which was POV.  That's not possible if the POV is a Layout.

Generally, we should put as much of our functionality into Animations as possible, is the conclusion of this, because we can switch them in real time.

rec:
Well, looking at it in the cold hard light of morning, I agree it's sorta feature creepy but I'm going to push it anyway because I really want to charge forward and fix these three outstanding issues... :-/

rec:
And what better time to push creepy code than Halloween?

adammhaile:
lol... fair enough

PR #1055, opened 2018-11-19
v3.4.34

adammhaile:
Fixing setup.py for windows


adammhaile:
@rec Just a hot fix. Turns out you broke setup.py for Windows about a month ago... https://github.com/ManiacalLabs/BiblioPixel/commit/ce8c34b16a93cdb63811e67d3bf4edb1780437e6#diff-2eeaed663bd0d25b7e608891384b7298
Never noticed since I don't really ever pip install it.
In setup.py you cannot initialize variables after you need them on Windows. No idea why, just that it's always been the case.

PR #1059, opened 2018-11-20
Move bibliopixel.util.colors back to bibliopixel.colors

rec:
The move of bibliopixel.colors to bibliopixel.util.colors was a mistake!

It happened when many other utilities got moved to `util/` which was generally good, but that one was wrong - there is just too much in `color/`, it's a natural top-level directory.

This change is backward and forward compatible, even if people are using undocumented features.

Adding @adammhaile just so I can say, "Oops!"  :-D


rec:
All right, I'm going to push this because I'm in a clean-up frenzy as we get thisclose to 3.5.  :-D

Back to documentation this afternoon...

PR #1081, opened 2018-11-27
Fix up documentation index

rec:


rec:
This is tiny, but I want to push it fast because the documentation index is confusing right now.

adammhaile:
+1

Issue #1084, opened 2018-11-28
Breakage in a non-project


adammhaile:
Ok.... so recent changes have made it so that you absolutely CANNOT run anything that's not a "project".
This seems to be what completely screws it up:
https://github.com/ManiacalLabs/BiblioPixel/blob/ea9c256417b56d074b0e3bec2a046a983f2264a1/bibliopixel/project/project.py#L59

That is only set if it's running a project. Then you can see here that `push_to_driver()` is only called if top_level == True:
https://github.com/ManiacalLabs/BiblioPixel/blob/f65695daa44545ce462d16f3a4b720b0854962a0/bibliopixel/animation/animation.py#L159

This needs to be fixed ASAP. We aren't supposed to break support for old scripts.

adammhaile:
Broken in this commit: https://github.com/ManiacalLabs/BiblioPixel/commit/c5edfe31aeeef8d9a37734c730827eb5f9c24d80

rec:
Ah, wrong, it's in another script.  

And running that test works!

rec:
Ah, so the stated issue is not the actual issue

> That [`Animation.top_levelo`] is only set if it's running a project. 

This is not so:  https://github.com/ManiacalLabs/BiblioPixel/blob/master/bibliopixel/animation/animation.py#L9

Because of backwards compatibility, and because people aren't necessarily calling the parent construction, when I add a new variable these days, it's almost always added that way - setting it as a static member in the parent class which the derived child class can override.

And I can run the two main programs I have here (the one checked in and some old one I have lying around).

So :-/ I don't have a way to reproduce this issue.  Can you find me one?

rec:
Or even better - a stack trace!

adammhaile:
See #1083

adammhaile:
There is no stacktrace it doesn't bomb out, it just never updates the driver.

rec:
I saw #1083.  Isn't that obviously a windows thing?

From that trace:

> WARNING - load - This "path" entry does not exist: "G"

And you're using drive G: elsewhere.

I'm not sure there are tests for `path` using Windows directory prefixes.  

I had certainly not thought about them, and `:` might even hit a wrong pattern match (e.g. "thinks it's a URL" or "thinks it's YAML".


----

Can you reproduce it on something that isn't the serial port?

I did try all of this on the test board before I pushed the release!

adammhaile:
Why would that be a windows thing?
[...]

Issue #1076, opened 2018-11-27
Some individual release_tests don't terminate


rec:
I thought I'd broken the signal handling, but I just tried both `bp restart` and `bp shutdown` and they seemed to work fine.

This is somewhat low priority because no one other than myself will likely see this, but it's annoying to me anyway.  :-D

PR #1092, opened 2018-11-30
v3.4.40

### LINKED ISSUE #1088: Matrix.py - ValueError: width * height cannot exceed total pixel count!
I'm trying to get the matrix.py (from https://github.com/ManiacalLabs/AllPixel/wiki/Matrix-Example) working but I get this error:

'Traceback (most recent call last):
File "matrix.py", line 16, in 
serpentine = True)
File "/usr/local/lib/python3.5/dist-packages/bibliopixel/layout/matrix.py", line 67, in init
% (self.width, self.height, self.numLEDs))
ValueError: width * height cannot exceed total pixel count! 32 * 32 > 9'

The only line in the example code I changed was:
'driver = Serial(num = 3*3, ledtype = LEDTYPE.WS2812)'
(using a 3x3 Nulsom ns-led-06 led matrix)

### LINKED ISSUE #1094: Remote animation does not shut down
You have to kill the process and sometimes `sudo lsof -t -i tcp:1337 | xargs kill -9`

I went back almost a year and the same behavior was there.

It has something to do with flask running in a separate process.  We now know we don't have to run flask that way, but I should be able to fix it by using the same idea as in the REST control, just in a separate process.

rec:
* Fix #1094, #1088, #1084, #1076 and probably #1085
* Make all Yaml dumping safe


rec:
I think we've resolved that.  I still need to run the release_tests from start to finish.

rec:
All right, I just ran all the tests and everything works - I'm going to push it and release it.

PR #1115, opened 2019-01-05
Script reorganization

rec:
@adammhaile, I'd just like you to look at the single "reorganize scripts" commit - the rest are uncontroversial.

There were getting to be too many scripts in that directory (and more coming) so I moved things around a bit, and I think I fixed everything, but in my experience "moving around things that are deployed" is a classic source of bugs, so if you could take a quickie to see if I haven't broken anything...

This is all part of my "onboard a new developer" project.  Tomorrow evening I'll be going through the materials with him...


adammhaile:
Will look tonight. Family is here now

PR #1127, opened 2019-01-08
Cache devices by hardware ID

rec:


rec:
I tested this on the 12k, which is my goto hardware at this moment for multidevices.

However, as we know this didn't show the problem before.  At least it shows I didn't make it worse!

adammhaile:
Ack, will test tonight

rec:
So I have no idea why we are getting these failures - they're ubiquitous on new branches, nothing to do with this branch

The root error is this:

    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (pytest 3.0.7 (/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages), Requirement.parse('pytest>=3.6'))

But we don't specify a version of pytest at all.  And AFAIK none of our dependencies even specify pytest as a dependency!

adammhaile:
Where do you see that?

rec:
Ah, forgot to paste this: https://travis-ci.org/ManiacalLabs/BiblioPixel/jobs/476942556

Note the "all checks failed" on this page.

adammhaile:
This might be helpful? https://github.com/Lukasa/hyper/issues/328

adammhaile:
@rec - Confirmed that works :D

rec:
OK, so how do we fix this CI issue then?

I tried all sorts of things, without success.

adammhaile:
I'm not really sure. I honestly don't get why it's only now happening...
I'll look more in a bit.

adammhaile:
Disable the coverage test for now and run it locally? That's what's
failing...

PR #1129, opened 2019-01-08
Require pytest 2.7

rec:


rec:
This failed - I'm going to try some variations...

rec:
OK - I disabled 3.5, but I added 3.7(!) so we are actually doing about the same.

I'm going to push this because it's gating pretty well everything else, and then we can deal with 3.5 later.

My theory is that others will run into the same issue in the coming weeks and months and someone else will figure it out.

PR #1146, opened 2019-11-25
Use mido multi_receive instead of MultiPort to receive midi messages

VonHirsch:
I tested on Raspberry Pi & Windows and in both cases MultiPort would not yield midi messages, however multi_receive works perfectly.

p.s. I would open up an issue on mido but the dev seems to be inactive there.


VonHirsch:
Failed due to FakeMido, should I also update that class and re-submit the PR?

adammhaile:
@VonHirsch Just update the code and push to your branch. I recommend `git commit --amend --no-edit; git push -f origin MidiFix` so there's no new commit.

VonHirsch:
ok thanks for that, done

Issue #1152, opened 2019-12-31
Poor performance with numpy


iceboundflame:
Enabling numpy with "numbers: float" causes a huge slowdown:
```
python: FRAME - animation - 15ms/63fps / Frame: 2ms / Update: 13ms
numpy (float): FRAME - animation - 61ms/16fps / Frame: 5ms / Update: 55ms
```

With `mtprof` I saw a ton of time spent in driver_base `_render()`. The following hack/proof of concept gets updates down to 4ms: https://github.com/iceboundflame/BiblioPixel/commit/325ab8fa336427e7d53055df81cbcc8ebe667067

Other important functionality such as palettes, matrix layout lookups are also Python-based, which makes them too slow for big arrays. To work around that, I have to directly manipulate `layout.color_list` using NumPy operations. (This is with an array of ~2300 LEDs on a Raspberry Pi, so any Python-based looping is an immediate frame rate killer.)

adammhaile:
yeah.... the whole float thing was an interesting concept but never really worked out or ever used much.
Your change looks reasonable. I'd definitely be OK with a PR for that (once the TODOs have been done).
Also - email me if you wouldn't mind: adammhaile@gmail.com
Would like to discuss some stuff re: BiblioPixel with you if you have time.

iceboundflame:
Sure! Here's a draft PR.

PR #1153, opened 2020-01-02
Speed up LED driver updates when using NumPy

iceboundflame:
Addresses #1152

Profiling revealed large time spent in `DriverBase._render()` method.
- Add a numpy path for `DriverBase._render()` to avoid Python-based looping.
- Add `Gamma.np_table` and `Gamma.batch_correct`.
- Add `Palette.np_table` and `Palette.batch_apply_palette`.

One concern: do we care that the package runs without `numpy` installed? Because importing `numpy` prevents non-numpy users from running Bibliopixel.

### DIFF
--- a/bibliopixel/colors/gamma.py
+++ b/bibliopixel/colors/gamma.py
@@ -1,3 +1,6 @@
+import numpy as np
+
+
 class Gamma(object):
     """
     Compute a fixed gamma table with 256 entries.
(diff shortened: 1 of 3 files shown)


iceboundflame:
One question: I'm not sure how threading is supposed to work - looks like there's a lock on brightness but not on _buf? Is threading actually used in the drivers or is it WIP?

